PR #3, opened 2016-12-13
Adding access to the SplitNames functionality

HJewkes:


CHANGES.txt
```diff
@@ -1,16 +1,19 @@
 CHANGES
+1.1.1
```
patricioe: 1.1.1 already exists:

http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22io.split.client%22%20AND%20a%3A%22java-client%22

HJewkes:
Ahh that makes sense - I didn't find a staging branch and was surprised

PR #4, opened 2016-12-14
Adding access to the SplitNames functionality

HJewkes:
Jira Ticket: [SPLIT-2315](https://splitio.atlassian.net/browse/SPLIT-2315)

### DIFF
--- a/src/main/java/io/split/client/SplitManager.java
+++ b/src/main/java/io/split/client/SplitManager.java
@@ -17,14 +17,14 @@ public interface SplitManager {
     List<SplitView> splits();
 
     /**
-     * Returns the features (or Split) registered with the SDK of this name.
+     * Returns the feature (or Split) registered with the SDK of this name.
      *
      * @return SplitView or null
      */
     SplitView split(String featureName);
 
     /**
-     * Returns the names of features (or Split) registered with the SDK.
+     * Returns the names of features (or Splits) registered with the SDK.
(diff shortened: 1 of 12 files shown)


src/main/java/io/split/client/SplitManager.java
```diff
@@ -22,4 +22,11 @@
      * @return SplitView or null
      */
     SplitView split(String featureName);
+
+    /**
+     * Returns the names of features (or Split) registered with the SDK.
```
SenhorCastor: typo Splits

HJewkes: Fixed

src/main/java/io/split/engine/matchers/LessThanOrEqualToMatcher.java
```diff
@@ -53,7 +53,7 @@ public String toString() {
     @Override
     public int hashCode() {
         int result = 17;
-        result = 31 * result + Long.hashCode(_compareTo);
+        result = 31 * result + (int)(_compareTo ^ (_compareTo >>> 32));
```
patricioe: is this better than Long.hashCode?

HJewkes: According to @SenhorCastor it was a Java 7 Compatability change

PR #7, opened 2016-12-16
Updating Hashing function to Murmur3

HJewkes:


src/main/java/io/split/client/utils/MurmurHash3.java
```diff
@@ -0,0 +1,338 @@
...
+
+    /**
+     * Returns the MurmurHash3_x86_32 hash.
+     */
+    public static int murmurhash3_x86_32(byte[] data, int offset, int len, int seed) {
+
+        final int c1 = 0xcc9e2d51;
+        final int c2 = 0x1b873593;
+
+        int h1 = seed;
```
patricioe: @adilaijaz the seed should be used at the very end correct?

adilaijaz:
Pato, where the seed gets used depends on the hashing algorithm. In murmur,
it gets used in the beginning.

client/src/main/java/io/split/engine/splitter/Splitter.java
```diff
@@ -11,8 +12,10 @@
  * @author adil
  */
 public class Splitter {
+    private static final int ALGO_MURMUR = 1;
```
patricioe: murmur is actually #2. see ticket: https://splitio.atlassian.net/browse/SPLIT-2578

patricioe: yeah confirmed. This is not correct.

HJewkes:
LGTM

PR #8, opened 2016-12-30
Murmur3 unsigned

patricioe:
@adilaijaz this PR branches off @HJewkes `murmur3` branch. Merging this means it will get merged to `murmur3` branch.


src/main/java/io/split/client/utils/MurmurHash3.java
```diff
@@ -111,7 +111,7 @@ public static int murmurhash3_x86_32(byte[] data, int offset, int len, int seed)
         h1 *= 0xc2b2ae35;
         h1 ^= h1 >>> 16;
 
-        return h1;
+        return h1 & 0xFFFFFFFFL;
```
adilaijaz: I'd recommend dropping a comment here explaining why the ANDing is happening

PR #16, opened 2017-02-28
Bumping Sub-Module Version

HJewkes:


client/src/main/java/io/split/client/LocalhostSplitFile.java
```diff
@@ -0,0 +1,130 @@
...
+        Map<String, String> onSplits = Maps.newHashMap();
+
+        try (BufferedReader reader = new BufferedReader(new FileReader(_file))) {
+            for (String line = reader.readLine(); line != null; line = reader.readLine()) {
+                line = line.trim();
+                if (line.isEmpty() || line.startsWith("#")) {
+                    continue;
+                }
+
+                String[] feature_treatment = line.split("\\s+");
```
sande: just a personal preference, I would have a POJO with fields feature, treatment rather than this array, but it was already there, so up to you.

client/src/main/java/io/split/client/LocalhostSplitFile.java
```diff
@@ -0,0 +1,130 @@
...
+
+                onSplits.put(feature_treatment[0], feature_treatment[1]);
+                _log.info("100% of keys will see " + feature_treatment[1] + " for " + feature_treatment[0]);
+            }
+        } catch (FileNotFoundException e) {
+            _log.warn("There was no file named " + _file.getPath() + " found. " +
+                    "We created a split client that returns default treatments for all features for all of your users. " +
+                    "If you wish to return a specific treatment for a feature, enter the name of that feature name and " +
+                    "treatment name separated by whitespace in " + _file.getPath() +
+                    "; one pair per line. Empty lines or lines starting with '#' are considered comments", e);
```
sande: Maybe add that lines with more than two words? are also ignored? (again up to you)

PR #17, opened 2017-03-01
Identify SDK

HJewkes:
```java

SplitAPIClientConfig config = SplitAPIClientConfig.builder()
  .publicApiUrl()
  .build();

SplitAPIClient apiClient = SplitApiClientBuilder.client("YOUR_ADMIN_API_KEY", config);

TrafficType type = apiClient.trafficTypes().get("User");
List<Attribute> attributes = apiClient.attributes().list(type);
Environment env = apiClient.environments().get("Production");
apiClient.identities().create(Identity.builder()
  .trafficType(type)
  .environment(env)
  .key("email@address.com")
  .addValue("attribute", "value")
  .addValue("number", 1000)
  .addValue("boolean", true)
  .build()
);
```

Open questions:
- Do we include this back in the Split Module?
- Do we use a Factory pattern?
- Do we use a different configuration object?
- Do we use the segmented client format (ie `apiClient.trafficTypes().list()`) or a flat client (ie `apiClient.listTrafficTypes()`)


api-client/src/main/java/io/split/api/SplitApiClientBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+//                .build();
+//
+//        URI rootTarget = URI.create(config.endpoint());
+//        URI eventsRootTarget = URI.create(config.eventsEndpoint());
+//
+
+        return new ApiClient();
+    }
+
+    public static SplitApiClient local() throws IOException {
```
HJewkes: Remove Local client (unneeded)

api-client/src/main/java/io/split/api/client/api/ApiAttributeClient.java
```diff
@@ -0,0 +1,32 @@
...
+public class ApiAttributeClient implements AttributeClient {
+    public List<Attribute> list(String trafficTypeId) {
+        return new ArrayList<>();
+    }
+
+    public List<Attribute> list(TrafficType trafficType) {
+        return list(trafficType.id());
+    }
+
+    public ApiAttributeClient create(Attribute attribute) throws IllegalArgumentException {
```
jwaimann: @HJewkes Question here. Why does it return a result of the same type as the service? Shouldn't it be a Boolean, to know if the action was successful or not, or at least the type of the object being created (I mean: Attribute)?

HJewkes: @patricioe requested creating a fluent api, so this is designed to allow for fluent execution of calls. He may have only been referring to the builder pattern for objects tho - so I'll let him clarify

patricioe: we decided to return a result after each execution method.

PR #18, opened 2017-03-23
Traffic Allocations for Twilio

adilaijaz:
- Support for Traffic Allocations
- Introduced an ImpressionListener so we do not have to expose the label in the getTreatment.
- We no longer have the ability to set refresh rates to less than 30 seconds.


client/src/main/java/io/split/client/SplitClientConfig.java
```diff
@@ -307,16 +354,20 @@ public Builder ready(int milliseconds) {
 
 
         public SplitClientConfig build() {
-            if (_featuresRefreshRate <= 0) {
-                throw new IllegalArgumentException("featuresRefreshRate must be > 0: " + _featuresRefreshRate);
+            if (_featuresRefreshRate < 30 ) {
+                throw new IllegalArgumentException("featuresRefreshRate must be >= 30: " + _featuresRefreshRate);
```
patricioe: should we not error out and set our default + logging WARN?

client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -145,7 +147,31 @@ private TreatmentLabelAndChangeNumber getTreatment(String matchingKey, String bu
...
+
+                if (!inRollout && parsedCondition.conditionType() == ConditionType.ROLLOUT) {
+
+                    if (parsedSplit.trafficAllocation() < 100) {
+                        // if the traffic allocation is 100%, no need to do anything special.
+                        int bucket = Splitter.getBucket(bucketingKey, parsedSplit.trafficAllocationSeed());
+
+                        if (bucket >= parsedSplit.trafficAllocation()) {
+                            // out of split
+                            return new TreatmentLabelAndChangeNumber(parsedSplit.defaultTreatment(), "not in split", parsedSplit.changeNumber());
```
patricioe: Can this `"not in split"` be a constant in a LABELs class or enum like other SDKs?
Same for `"killed"`, `"exception"` etc.

client/src/main/java/io/split/client/SplitFactoryBuilder.java
```diff
@@ -113,7 +116,18 @@ public static synchronized SplitFactory build(String apiToken, SplitClientConfig
         final RefreshableSplitFetcherProvider splitFetcherProvider = new RefreshableSplitFetcherProvider(splitChangeFetcher, splitParser, findPollingPeriod(RANDOM, config.featuresRefreshRate()), gates);
 
         // Impressions
-        final TreatmentLog treatmentLog = ImpressionsManager.instance(httpclient, config);
+        final ImpressionsManager splitImpressionListener = ImpressionsManager.instance(httpclient, config);
+        final ImpressionListener impressionListener;
+
+        if (config.impressionListener() != null) {
+            AsynchronousImpressionListener wrapper = AsynchronousImpressionListener.build(config.impressionListener(), config.impressionListenerCapactity());
```
patricioe: in this branch of the if statement (when both listeners exist), which one executes first?

client/src/main/java/io/split/client/impressions/Impression.java
```diff
@@ -0,0 +1,54 @@
...
+    private final String _key;
+    private final String _bucketingKey;
+    private final String _split;
+    private final String _treatment;
+    private final long _time;
+    private final String _appliedRule;
+    private final Long _changeNumber;
+
+
+    public Impression(String key, String bucketingKey, String split, String treatment, long time, String appliedRule, Long changeNumber) {
```
patricioe: Don't we already have KeyImpression?

client/src/main/java/io/split/client/SplitClientConfig.java
adilaijaz: tell me what you want and I can do that

client/src/main/java/io/split/client/SplitFactoryBuilder.java
adilaijaz: when both listeners exist, then we create a federated listener of our listener (that uploads impressions to our endpoint) and theirs

client/src/main/java/io/split/client/impressions/Impression.java
adilaijaz: we do, but that is an internal API. This is an API meant for external consumption. Reasons for keeping them separate:

- We may be in a world down the road where we are sending more data to ourselves than to an external API
- The internal api says 'label' and 'feature'. For external API, i am using 'appliedRule' and 'spilt' to be more consistent with the UI.

client/src/test/java/io/split/engine/splitter/SplitterTest.java
```diff
@@ -97,6 +97,11 @@ public void ifHundredPercentOneTreatmentWeShortcut() {
         assertThat(Splitter.getTreatment("13", 15, partitions), is(equalTo("on")));
     }
 
+    @Test
+    public void getBucket() {
+        System.out.println(Splitter.getBucket("pato@split.io", 123));
```
patricioe: can we improve this?

pom.xml
```diff
@@ -1,145 +1,144 @@
...
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>1.7</maven.compiler.source>
-        <maven.compiler.target>1.7</maven.compiler.target>
-    </properties>
-    <modules>
-        <module>testing</module>
-        <module>client</module>
-    </modules>
-    <build>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
patricioe: what changed here? more dependencies?

client/src/main/java/io/split/client/SplitClientConfig.java
```diff
@@ -60,6 +66,8 @@ private SplitClientConfig(String endpoint,
         _ready = ready;
         _debugEnabled = debugEnabled;
         _labelsEnabled = labelsEnabled;
+        _impressionListener = impressionListener;
+        _impressionListenerCapacity = impressionListenerCapacity;
```
patricioe: what's the default?

client/src/main/java/io/split/client/impressions/AsynchronousImpressionListener.java
```diff
@@ -0,0 +1,68 @@
...
+    private final ImpressionListener _delegate;
+    private final ExecutorService _executor;
+
+    public static AsynchronousImpressionListener build(ImpressionListener delegate, int capacity) {
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setDaemon(true)
+                .setNameFormat("impression-listener-wrapper-%d")
+                .build();
+
+        ExecutorService executor = new ThreadPoolExecutor(2, 2, 0L, TimeUnit.MILLISECONDS, new ArrayBlockingQueue<Runnable>(capacity), threadFactory);
```
patricioe: What's the rejection policy used?

client/src/main/java/io/split/client/SplitClientConfig.java
adilaijaz: There is no default. The method in the builder is:

public Builder impressionListener(ImpressionListener impressionListener, int capacity) {
  // code
}

which means that when you provide the impression listener, you also provide the capacity

client/src/main/java/io/split/client/impressions/AsynchronousImpressionListener.java
adilaijaz: AbortPolicy. The wrapper catches any RejectedExecutions.

client/src/test/java/io/split/engine/splitter/SplitterTest.java
adilaijaz: fixed

pom.xml
adilaijaz: just the version number

PR #19, opened 2017-03-31
Fix

adilaijaz:


patricioe:
reviewing it. Please make sure you issue against development branch going forward.

client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -65,6 +65,11 @@ public String getTreatment(Key key, String split, Map<String, Object> attributes
             return Treatments.CONTROL;
         }
 
+        if (key.matchingKey() == null || key.bucketingKey() == null) {
```
patricioe: I'm not sure I follow this. Did you mean `and` ?

patricioe: I'm not sure we should return control in this case. I believe people can still call us with `Key` and only fill up matching key. thoughts?

PR #22, opened 2017-05-04
Impression Metadata into local impression listener

adilaijaz:


client/NEWS.txt
```diff
@@ -1,5 +1,11 @@
 NEWS
 
+2.0.2
+- Attributes are send to local impression listeners
```
HJewkes: `s/are send/can be sent/`

SenhorCastor: think it is "are sent". check with a native speaker

client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -120,9 +121,10 @@ private String getTreatment(String matchingKey, String bucketingKey, String spli
         }
     }
 
-    private void recordStats(String matchingKey, String bucketingKey, String split, long start, String result, String operation, String label, Long changeNumber) {
+    private void recordStats(String matchingKey, String bucketingKey, String split, long start, String result,
+                             String operation, String label, Long changeNumber, Map<String, Object> impressionMetadata) {
```
HJewkes: `s/impressionMetadata/attributes/`

client/src/main/java/io/split/client/impressions/Impression.java
```diff
@@ -12,16 +14,18 @@
     private final long _time;
     private final String _appliedRule;
     private final Long _changeNumber;
+    private final Map<String, Object> _impressionMetadata;
```
HJewkes: Change ImpressionMetadata to Attributes throughout this code

HJewkes:
I would strongly advocate for adding a boolean value to the configuration that enables this functionality, and defaults to not doing so. Primarily, this by default preserves our contract with customers to never send on attributes past the get treatment check. Customers may be sending this impression data straight to an endpoint, queue or database and be surprised that the people on the other end have access to that data. Also, sending the attribute data to the impressions listener implies to customers that this data is on the impressions we are receiving. We will certainly document that that is not the case, but by requiring a configuration to enable this it ensures that customers have engaged with that documentation before the data starts propagating.
@patricioe @adilaijaz

PR #23, opened 2017-05-05
improvements for runtime on new matchers

patricioe:


client/src/main/java/io/split/engine/matchers/collections/PartOfSetMatcher.java
```diff
@@ -33,13 +31,11 @@ public boolean match(Object key) {
             return false;
         }
 
-        Set<String> keyAsSet = toSetOfStrings((Collection) key);
```
HJewkes: Is toSetOfStrings() still used elsewhere or can that be removed now?

PR #24, opened 2017-06-02
Exposing destroy functionality to support shutting down background th…

HJewkes:
…reads without JVM shutdown


client/src/main/java/io/split/client/SplitFactoryBuilder.java
```diff
@@ -265,4 +269,11 @@ public static void main(String... args) throws IOException, InterruptedException
             _log.error(io.getMessage(), io);
         }
     }
+
+    public static void destroy() {
+        for (int i = 0; i < DESTROY_FUNCTIONS.size(); i++) {
+            DESTROY_FUNCTIONS.get(i).run();
+            DESTROY_FUNCTIONS.remove(i);
+        }
+    }
```
patricioe: why don't we have a `client.destroy()`?

PR #25, opened 2017-06-09
Fix root POM to not implicitly introduce compile-scoped JUnit dependency

JoshRosen:
The current Java SDK ends up pulling in JUnit as a `compile`-scoped dependency. To see this, create a folder with the POM

```
<?xml version="1.0" encoding="UTF-8"?>
<project>
  <groupId>com.example</groupId>
  <artifactId>mylib</artifactId>
  <version>1.0</version>
  <dependencies>
    <dependency>
        <version>2.0.2</version>
    </dependency>
  </dependencies>
</project>
```

and then run ` mvn dependency:tree`:

```
[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ mylib ---
[INFO] com.example:mylib:jar:1.0
[INFO] \- io.split.client:java-client:jar:2.0.2:compile
[INFO]    +- com.google.guava:guava:jar:18.0:compile
[INFO]    +- org.slf4j:slf4j-api:jar:1.7.12:compile
[INFO]    +- commons-logging:commons-logging:jar:1.2:compile
[INFO]    +- commons-codec:commons-codec:jar:1.9:compile
[INFO]    \- junit:junit:jar:4.12:compile
[INFO]       \- org.hamcrest:hamcrest-core:jar:1.3:compile
```

The problem here is that the root POM is using `<dependencies>` instead of `<dependencyManagement>`. Here, I've gone ahead and updated the POM to use `<dependencyManagement>` so you can manage versions centrally in the root POM but avoid adding dependencies to all subprojects. I manually verified that this fix removes the `compile`-scoped JUnit dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,18 +5,20 @@
     <groupId>io.split.client</groupId>
     <artifactId>java-client-parent</artifactId>
     <version>2.0.3-rc2</version>
-    <dependencies>
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.12</version>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-            <version>1.7.21</version>
-        </dependency>
-    </dependencies>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>junit</groupId>
+                <artifactId>junit</artifactId>
+                <version>4.12</version>
+            </dependency>
+            <dependency>
+                <groupId>org.slf4j</groupId>
+                <artifactId>slf4j-api</artifactId>
+                <version>1.7.12</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
     <packaging>pom</packaging>
     <name>Java Client</name>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -5,18 +5,20 @@
...
-    <dependencies>
-        <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
-            <version>4.12</version>
-        </dependency>
-        <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-api</artifactId>
-            <version>1.7.21</version>
```
JoshRosen: The root POM was declaring this version, but both children were using the older `1.7.12` (the child POM takes precedence). To avoid a version upgrade as part of this change I decided to downgrade in the root POM to match the version that was actually being used. Feel free to edit if you'd like to bump.

patricioe:
Sorry about that. It was an oversight.

We'll fix ASAP.

JoshRosen:
No worries; our dependency management system is really really strict which is why it caught this problem. For now I've just added an exclusion, but I figured I'd submit a PR upstream to fix this there, too.

patricioe:
@JoshRosen merged. Thanks!

PR #26, opened 2017-07-10
Adding dependency matcher logic

HJewkes:


client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -94,30 +95,21 @@ private String getTreatment(String matchingKey, String bucketingKey, String spli
...
-                        result._treatment,
-                        "sdk.getTreatment",
-                        _config.labelsEnabled() ? result._label : null,
-                        result._changeNumber,
-                        attributes);
-            }
+            TreatmentLabelAndChangeNumber result = getTreatmentResultWithoutImpressions(matchingKey, bucketingKey, split, attributes);
 
-            return result._treatment;
+            recordStats(
```
HJewkes: Fair question. I can change the catch(Exception) to catch throwable - since that seems to be the intention anyway

client/src/main/java/io/split/engine/matchers/DependencyMatcher.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        String result = splitClient.getTreatmentWithoutImpressions(
+                (String) matchValue,
+                bucketingKey,
+                _split,
+                attributes
+        );
+
+//        if(Treatments.isControl(result)) {
```
HJewkes: I personally believe that we should return control if the parent does, and implemented accordingly. We made a call this morning to not do that, but I figured for now I'd comment out the code rather than fully ripping it out in case we change back.

HJewkes:
Always happy to get more eyeballs. The changes to Matcher.java interface are in accordance with the agreed upon spec with a small deviation of passing in matchingValue and bucketingKey as two separate values rather than a combined Key object due to the way we're doing typing magic in Java

client/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.split.client</groupId>
         <artifactId>java-client-parent</artifactId>
-        <version>2.0.4</version>
+        <version>2.0.5</version>
```
patricioe: make it 2.0.5-RC1 please

client/src/main/java/io/split/client/dtos/MatcherType.java
```diff
@@ -8,20 +8,23 @@
...
 
-    /**String Matchers **/
+    /* String Matcher */
     STARTS_WITH,
     ENDS_WITH,
-    CONTAINS_STRING
+    CONTAINS_STRING,
+
+    /* Dependency Matcher */
+    IN_SPLIT_TREATMENT
```
patricioe: are we missing boolean or boolean matcher has not be implemented yet?

client/src/main/java/io/split/engine/matchers/DependencyMatcher.java
patricioe: also because returning control here will cause the dependent split to return control, thus, aborting the rest of the evaluation chain in the dependent.

I decided it was best to just not match instead of returning CONTROL.

pom.xml
```diff
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>io.split.client</groupId>
     <artifactId>java-client-parent</artifactId>
-    <version>2.0.4</version>
+    <version>2.0.5</version>
```
patricioe: 2.0.5-RC1

testing/pom.xml
```diff
@@ -6,7 +6,7 @@
     <parent>
         <groupId>io.split.client</groupId>
         <artifactId>java-client-parent</artifactId>
-        <version>2.0.4</version>
+        <version>2.0.5</version>
```
patricioe: 2.0.5-RC1

patricioe:
@SenhorCastor your review is still marked as pending.

@HJewkes I suggested to release it under 2.0.5-rc1 so we can deploy in testing apps and monitor DD and QoS for a day or two.

PR #27, opened 2017-07-10
Boolean matcher

HJewkes:


client/src/main/java/io/split/engine/matchers/Transformers.java
```diff
@@ -58,6 +60,24 @@ public static Long asDateHourMinute(Object obj) {
...
+        if (obj == null) {
+            return null;
+        }
+
+        if (obj instanceof Boolean) {
+            return (Boolean) obj;
+        }
+
+        if (obj instanceof String) {
+            if (Sets.newHashSet("true", "false").contains(((String) obj).toLowerCase())) {
```
SenhorCastor: have a static set so its not initialized every time?, same with the (String obj)

client/src/main/java/io/split/engine/matchers/BooleanMatcher.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public boolean match(Object matchValue, String bucketingKey, Map<String, Object> attributes, SplitClientImpl splitClient) {
+        if (matchValue == null) {
+            return false;
+        }
+
+        Boolean valueAsBoolean = asBoolean(matchValue);
+
+        return valueAsBoolean != null && valueAsBoolean == _booleanValue;
```
adilaijaz: valueAsBoolean.booleanValue() == _booleanValue

client/src/main/java/io/split/engine/matchers/BooleanMatcher.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        BooleanMatcher that = (BooleanMatcher) o;
+
+        return _booleanValue == that._booleanValue;
```
adilaijaz: same here. Wouldn't we want to do a .booleanValue() here.

HJewkes: Unboxing for equality checks is unnecessary, Java does it automatically: https://stackoverflow.com/questions/30454635/when-using-for-a-primitive-and-a-boxed-value-is-autoboxing-done-or-is-unbox

client/src/test/java/io/split/engine/matchers/AttributeMatcherTest.java
```diff
@@ -54,6 +54,17 @@ public void works_less_than_or_equal_to() {
     }
 
     @Test
+    public void works_boolean() {
+        AttributeMatcher matcher = new AttributeMatcher("value", new BooleanMatcher(true), false);
+        assertThat(matcher.match("ignore", null, ImmutableMap.<String, Object>of("value", true), null), is(true));
+        assertThat(matcher.match("ignore", null, ImmutableMap.<String, Object>of("value", "true"), null), is(true));
```
patricioe: could you please add a test with upper case as input?

patricioe:
Approved on my end. Left a comment to add some more tests.

PR #31, opened 2017-09-21
Proxy support and dependency split fix

patricioe:
Startup logs will look like this from now on:

```  
   INFO [main] (SplitFactoryImpl.java:95) - Initializing Split SDK with proxy settings <---- if proxy is used
   INFO [split-splitFetcher-0] (SDKReadinessGates.java:63) - splits are ready
   INFO [split-segmentFetcher-0] (SDKReadinessGates.java:112) - employees segment is ready
```


client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -80,14 +87,31 @@ public SplitFactoryImpl(String apiToken, SplitClientConfig config) throws IOExce
...
+                .addInterceptorLast(new GzipDecoderResponseInterceptor());
+
+        // Set up proxy is it exists
+        if (config.proxy() != null) {
+            _log.info("Initializing Split SDK with proxy settings");
+            DefaultProxyRoutePlanner routePlanner = new DefaultProxyRoutePlanner(config.proxy());
+            httpClientbuilder.setRoutePlanner(routePlanner);
+
+            if (config.proxyUsername() != null && config.proxyPassword() != null){
+                CredentialsProvider credsProvider = new BasicCredentialsProvider();
```
SenhorCastor: Should we log (in debug) something like "Initializaing proxy with user: userr and password: ***** 
just to make sure for debugging purposes to know that we are setting user/pass

PR #32, opened 2017-11-09
WIP for track() api

patricioe:


client/src/main/java/io/split/client/dtos/Event.java
```diff
@@ -0,0 +1,29 @@
+package io.split.client.dtos;
+
+import com.google.common.base.Objects;
+
+public class Event {
+
+    public String eventTypeId;
+    public String trafficTypeId;
```
patricioe: rename to `trafficTypeName`

client/src/main/java/io/split/client/EventClient.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        EventSenderTask(CloseableHttpClient httpclient, URI eventsTarget, List<Event> events) {
+            _client = httpclient;
+            _data = events;
+            _endpoint = eventsTarget;
+        }
+
+        @Override
+        public void run() {
```
patricioe: this can be templatized

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -97,4 +97,29 @@
...
+     * Enqueue a new event to be sent to split data collection services
+     *
+     * Example:
+     *      client.track(“account”, “Split Software”, “checkout”)
+     *
+     * @param trafficType the type of the event
+     * @param key the identifier of the entity
+     * @param eventType the type of the event
+     */
+    void track(String trafficType, String key, String eventType);
```
patricioe: @sande this needs to return a boolean

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -97,4 +97,29 @@
...
+     *
+     * Example:
+     *      client.track(“account”, “Split Software”, “checkout”, 200.00)
+     *
+     * @param trafficType the type of the event
+     * @param key the identifier of the entity
+     * @param eventType the type of the event
+     * @param value the value of the event
+     */
+    void track(String trafficType, String key, String eventType, double value);
```
patricioe: @sande this needs to return a boolean too

client/src/main/java/io/split/client/dtos/Event.java
sande: done

client/src/main/java/io/split/client/SplitClient.java
sande: done

client/src/main/java/io/split/client/SplitClient.java
sande: done.

client/src/main/java/io/split/client/EventClient.java
sande: done

PR #44, opened 2018-04-11
Whitelist overrides localhost

adilaijaz:
For Schlumberger.

See One Pager: https://docs.google.com/document/d/1asw7G41KVQ9R1F-2daE4RLkZ68p4GyjWhvx0YX67fk8/edit?usp=sharing


client/src/main/java/io/split/client/SplitAndKey.java
```diff
@@ -0,0 +1,67 @@
...
+    public String split() {
+        return _split;
+    }
+
+    public String key() {
+        return _key;
+    }
+
+    @Override
+    public boolean equals (Object o) {
```
SenhorCastor: nit extra space between equals and (

pom.xml
```diff
@@ -1,147 +1,146 @@
...
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>1.7</maven.compiler.source>
-        <maven.compiler.target>1.7</maven.compiler.target>
-    </properties>
-    <modules>
-        <module>testing</module>
-        <module>client</module>
-    </modules>
-    <build>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
SenhorCastor: same

client/CHANGES.txt
```diff
@@ -1,5 +1,8 @@
 CHANGES
 
+2.3.1
+- Added support for overrides in localhost mode
```
SenhorCastor: overrides of keys? key overrides?

client/src/main/java/io/split/client/LocalhostSplitClientAndFactory.java
```diff
@@ -0,0 +1,69 @@
...
+ * @author adil
+ */
+public final class LocalhostSplitClientAndFactory implements SplitClient {
+    private static final Logger _log = LoggerFactory.getLogger(LocalhostSplitClientAndFactory.class);
+
+    private LocalhostSplitFactory _factory;
+    private LocalhostSplitClient _splitClient;
+
+    public LocalhostSplitClientAndFactory(LocalhostSplitFactory container, LocalhostSplitClient client) {
+        _factory = container;
```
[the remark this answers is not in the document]
adilaijaz: Done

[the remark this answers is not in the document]
adilaijaz: Done

client/NEWS.txt
```diff
@@ -1,5 +1,8 @@
 NEWS
 
+2.3.1
+- Added support for key overrides in localhost mode
```
patricioe: so here in NEWS the idea is not to repeat what it's in CHANGES but so make it more of a sentences:

`- Adds the ability to include keys to the locamode file to use for matching responses to getTreatment`

client/src/main/java/io/split/client/LocalhostSplitClient.java
```diff
@@ -18,13 +19,13 @@
...
 public final class LocalhostSplitClient implements SplitClient {
-    private LocalhostSplitFactory _container;
-    private ImmutableMap<String, String> _featureToTreatmentMap;
+    private static final Logger _log = LoggerFactory.getLogger(LocalhostSplitClient.class);
 
-    public LocalhostSplitClient(LocalhostSplitFactory container, Map<String, String> featureToTreatmentMap) {
-        checkNotNull(featureToTreatmentMap, "featureToTreatmentMap must not be null");
-        _featureToTreatmentMap = ImmutableMap.copyOf(featureToTreatmentMap);
-        _container = container;
+    private Map<SplitAndKey, String> _map;
```
patricioe: can we name this anything other than map? for example: `_evaluationsOverride` ?

client/src/main/java/io/split/client/LocalhostSplitClientAndFactory.java
```diff
@@ -0,0 +1,72 @@
...
+        return _splitClient.getTreatment(key, split);
+    }
+
+    @Override
+    public String getTreatment(String key, String split, Map<String, Object> attributes) {
+        return _splitClient.getTreatment(key, split, attributes);
+    }
+
+    @Override
+    public String getTreatment(Key key, String split, Map<String, Object> attributes) {
```
patricioe: why are these methods are and also in the client?

adilaijaz: because this is a composite class meant to pull out the factory from the client. 

Before my changes, the LocalhostSplitClient had a reference to the factory that created it. That seems odd. Plus it makes unit testing harder. So I pulled the factory out of the client and made a composite class that could delegate to the client for client work and destroy the factory when it needs to.

client/src/main/java/io/split/client/LocalhostSplitClient.java
adilaijaz: it is not the overrides. It is the entire mapping read from .splits file

PR #59, opened 2018-12-13
fix race condition with loading events

patricioe:
``` INFO [split-segmentFetcher-0] (SDKReadinessGates.java:110) - Registered segment: demo_orgs
 INFO [split-splitFetcher-0] (SDKReadinessGates.java:62) - splits are ready
 INFO [split-segmentFetcher-0] (RefreshableSegment.java:135) - demo_orgs added keys: [2dfcc270-aacc-11e8-a8d2-xxxxxxxxxxx, 7479b040-aa14-11e8-91c4-xxxxxxxxxxx, 745622b0-aa14-11e8-91c4-12bc02531368... 5 others]
 INFO [split-segmentFetcher-0] (SDKReadinessGates.java:130) - demo_orgs segment is ready```

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.split.client</groupId>
         <artifactId>java-client-parent</artifactId>
-        <version>3.0.3</version>
+        <version>3.0.4</version>
     </parent>
     <artifactId>java-client</artifactId>
     <packaging>jar</packaging>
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: client/pom.xml [hunk 1]
```diff
@@ -156,5 +156,11 @@
             <version>1.10.19</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
SenhorCastor: Is this intended?

client/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,15 @@
+log4j.rootLogger=info, stdout
```
SenhorCastor: is this file intended\?

client/pom.xml
patricioe: yes, for testing. Otherwise there are no logs in the console.

client/src/test/resources/log4j.properties
patricioe: yes

PR #77, opened 2019-03-12
adding ability to check traffic type names on track calls

SenhorCastor:


client/src/main/java/io/split/engine/experiments/RefreshableSplitFetcher.java
```diff
@@ -85,12 +99,21 @@ public ParsedSplit fetch(String test) {
         return Lists.newArrayList(_concurrentMap.values());
     }
 
+    @Override
+    public Set<String> fetchUsedTrafficTypes() {
```
sande: would the name fetchKnownTrafficTypes() make more sense?

client/src/main/java/io/split/engine/experiments/RefreshableSplitFetcher.java
```diff
@@ -85,12 +99,21 @@ public ParsedSplit fetch(String test) {
...
+        // We return the "keys" of the multiset that have a count greater than 0
+        // If the multiset has [{"user",2}.{"account",0}], elementSet only returns
+        // ["user"] (it ignores "account")
+        return Sets.newHashSet(_concurrentTrafficTypeNameSet.elementSet());
```
sande: is it valid / worth making this unmodifiable?

client/src/main/java/io/split/engine/experiments/RefreshableSplitFetcher.java
```diff
@@ -160,21 +185,34 @@ public void runWithoutExceptionHandling() throws InterruptedException {
                 if (split.status != Status.ACTIVE) {
                     // archive.
                     toRemove.add(split.name);
+                    if (split.trafficTypeName != null) {
```
sande: Fer, I need more context here to understand why you're marking a traffic type for removal based on the split's status

client/src/main/java/io/split/engine/experiments/SplitFetcher.java
```diff
@@ -10,6 +11,16 @@
...
 
+    /**
+     * Fetches all the traffic types that are being used by the splits that are currently stored.
+     *
+     * For example, if the fetcher currently contains three splits, one of traffic type "account"
+     * and two of traffic type "user", this method will return ["account", "user"]
+     *
+     * @return a set of all the traffic types used by the parsed splits
+     */
+    Set<String> fetchUsedTrafficTypes();
```
sande: see other comment regarding naming

PR #83, opened 2019-04-03
Dynamic configs

patricioe:


client/src/main/java/io/split/client/api/SplitView.java
```diff
@@ -16,4 +14,5 @@
     public boolean killed;
     public List<String> treatments;
     public long changeNumber;
+    public Map<String, String> config;
```
mmelograno: This should be plural because it will have the configurations for all the treatments, so it should be `configs` instead of `config`

client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -106,42 +109,77 @@ public String getTreatment(Key key, String split, Map<String, Object> attributes
...
+
+    @Override
+    public SplitResult getTreatmentWithConfig(String key, String split, Map<String, Object> attributes) {
+        return getTreatmentWithConfigInternal(key, null, split, attributes);
+    }
+
+    @Override
+    public SplitResult getTreatmentWithConfig(Key key, String split, Map<String, Object> attributes) {
+        if (key == null) {
+            _log.error("getTreatment: you passed a null key, the key must be a non-empty string");
```
mmelograno: Is the word `getTreatment` prompted also when you call `getTreatmentWithConfig`? We decided to display `getTreatmentWithConfig` in the case that user call that method

client/src/main/java/io/split/client/LocalhostSplitFactory.java
```diff
@@ -19,26 +20,36 @@
...
         _log.info("home = " + directory);
 
-        _splitFile = new LocalhostSplitFile(this, directory, FILENAME);
+        if (directory == null) {
+            directory = "";
+        }
+
+        File yaml = new File(directory, file);
+
+        if (yaml.exists()) {
```
awgr: Should we test if the file is readable (permissions-wise) and log an error if it is not?

client/src/main/java/io/split/client/LocalhostSplitManager.java
```diff
@@ -23,28 +23,28 @@
...
         Map<String, Set<String>> splitsToTreatments = Maps.newHashMap();
-        for (Map.Entry<SplitAndKey, String> entry : splitAndKeyStringMap.entrySet()) {
+        for (Map.Entry<SplitAndKey, LocalhostSplit> entry : splitAndKeyStringMap.entrySet()) {
             String split = entry.getKey().split();
             if (!splitsToTreatments.containsKey(split)) {
                 splitsToTreatments.put(split, Sets.<String>newHashSet());
             }
             Set<String> treatments = splitsToTreatments.get(split);
-            treatments.add(entry.getValue());
+            treatments.add(entry.getValue().treatment);
```
awgr: `treatment` is a public field on `LocalhostSplit` ?

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -92,6 +93,49 @@
...
+     * to users created after a certain date.
+     *
+     * @param key         a unique key of your customer (e.g. user_id, user_email, account_id, etc.) MUST not be null.
+     * @param split    the feature we want to evaluate. MUST NOT be null.
+     * @return the evaluated treatment, the default treatment of this feature, or 'control'.
+     */
+    SplitResult getTreatmentWithConfig(String key, String split);
+
+    /**
+     * Same as {@link #getTreatment(String, String, Map)} but it returns the configuration associated to the
```
awgr: can we use an `{@link` tag for `the configuration`?

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -92,6 +93,49 @@
...
+     * matching treatment if any. Otherwise {@link SplitResult.configurations()} will be null.
+     * <p/>
+     * <p/>
+     * Examples include showing a different treatment to users on trial plan
```
awgr: An example usage here would be good - like a code snippet with the intended usage pattern

client/src/main/java/io/split/client/AlwaysReturnControlSplitClient.java
```diff
@@ -14,6 +15,8 @@
  */
 public class AlwaysReturnControlSplitClient implements SplitClient {
 
+    private static final SplitResult RESULT_CONTROL = new SplitResult(Treatments.CONTROL, null);
```
HJewkes: When a configuration is not defined, should we return an empty map instead of null? It would simplify / remove NPEs

client/src/main/java/io/split/client/LocalhostSplit.java
```diff
@@ -0,0 +1,20 @@
...
+    public String treatment;
+    public String config;
+
+    public LocalhostSplit(String treatment, String config) {
+        this.treatment = treatment;
+        this.config = config;
+    }
+
+    public static LocalhostSplit of(String treatment) {
+        return new LocalhostSplit(treatment, null);
```
HJewkes: When a configuration is not defined, should we return an empty map instead of null? It would simplify / remove NPEs

client/src/main/java/io/split/client/LocalhostSplitClient.java
```diff
@@ -20,47 +21,75 @@
  */
 public final class LocalhostSplitClient implements SplitClient {
     private static final Logger _log = LoggerFactory.getLogger(LocalhostSplitClient.class);
+    private static SplitResult SPLIT_RESULT_CONTROL = new SplitResult(Treatments.CONTROL, null);
```
HJewkes: When a configuration is not defined, should we return an empty map instead of null? It would simplify / remove NPEs

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -92,6 +93,49 @@
...
+     * vs. premium plan. Another example is to show a different treatment
+     * to users created after a certain date.
+     *
+     * @param key         a unique key of your customer (e.g. user_id, user_email, account_id, etc.) MUST not be null.
+     * @param split    the feature we want to evaluate. MUST NOT be null.
+     * @return the evaluated treatment, the default treatment of this feature, or 'control'.
```
HJewkes: Should we mention the configuration in the @return here

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -92,6 +93,49 @@
...
+     * <p/>
+     * <p/>
+     * Examples include showing a different treatment to users on trial plan
+     * vs. premium plan. Another example is to show a different treatment
+     * to users created after a certain date.
+     *
+     * @param key         a unique key of your customer (e.g. user_id, user_email, account_id, etc.) MUST not be null.
+     * @param split    the feature we want to evaluate. MUST NOT be null.
+     * @param attributes of the customer (user, account etc.) to use in evaluation. Can be null or empty.
+     * @return the evaluated treatment, the default treatment of this feature, or 'control'.
```
HJewkes: Should we mention the configuration in the @return here

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -92,6 +93,49 @@
...
+
+    /**
+     * Same as {@link #getTreatment(Key, String, Map)} but it returns the configuration associated to the
+     * matching treatment if any. Otherwise {@link SplitResult.configurations()} will be null.
+     *
+     * @param key the matching and bucketing keys. MUST NOT be null.
+     * @param split the feature we want to evaluate. MUST NOT be null.
+     * @param attributes of the entity (user, account etc.) to use in evaluation. Can be null or empty.
+     *
+     * @return the evaluated treatment, the default treatment of this feature, or 'control'.
```
HJewkes: Should we mention the configuration in the @return here

client/src/main/java/io/split/client/AlwaysReturnControlSplitClient.java
patricioe: `config` is a string.

HJewkes: 🤦‍♂️ Wow not sure how i missed that. Ok null seems fair then

PR #87, opened 2019-05-22
Feature/track properties

mredolatti:
- Allow Gson to serialize null values
- Add custom serializer for Doubles to avoid adding a floating point to every number

### DIFF
--- a/client/src/main/java/io/split/client/utils/Json.java
+++ b/client/src/main/java/io/split/client/utils/Json.java
@@ -17,9 +17,9 @@ public class Json {
     private static final Gson _json = new GsonBuilder()
             .serializeNulls()  // Send nulls
             .registerTypeAdapter(Double.class, new JsonSerializer<Double>() {
-                @Override
 
                 // Send integers as such
+                @Override
                 public JsonElement serialize(Double src, Type typeOfSrc, JsonSerializationContext context) {
                     if (src == src.longValue())
                         return new JsonPrimitive(src.longValue());
--- a/client/src/main/java/io/split/client/utils/Json.java
+++ b/client/src/main/java/io/split/client/utils/Json.java
@@ -1,13 +1,16 @@
 package io.split.client.utils;
 
 import com.google.gson.Gson;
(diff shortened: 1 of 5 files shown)


client/src/main/java/io/split/client/utils/Json.java
```diff
@@ -1,13 +1,32 @@
...
 
-    private static final Gson _json = new Gson();
+    private static final Gson _json = new GsonBuilder()
+            .serializeNulls()  // Send nulls
+            .registerTypeAdapter(Double.class, new JsonSerializer<Double>() {
+
+                // Send integers as such
+                @Override
+                public JsonElement serialize(Double src, Type typeOfSrc, JsonSerializationContext context) {
+                    if (src == src.longValue())
```
patricioe: @mredolatti this is not good. What are you trying to do here?

pom.xml
```diff
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>io.split.client</groupId>
     <artifactId>java-client-parent</artifactId>
-    <version>3.1.1</version>
+    <version>3.2.0</version>
```
patricioe: this is missing the other 2 pom files

mredolatti: Thank you for updating that @patricioe

client/src/main/java/io/split/client/utils/Json.java
mredolatti: Whenever Json parses an "Object" type, which happens to be a number, it serializes it as a Double, and adds a decimal point to it. so 123 becomes 123.0 (ref: https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java#L78), prior to sending it. This then becomes stringified in the BE, and it gets stored in elasticsearch as "123.0" . I'm not really sure how exp does the comparisons and whether this might be an issue or not, but our QoS fails for sure.

If we're 100% sure it won't be an issue for exp  we can always remove that and add extra logic in our QoS to test for equality under those conditions.

PR #92, opened 2019-07-10
Add support for optional timestamp parameter when tracking an event.

valesken:
We noticed while using the Java SDK that timestamps were being set automatically to `System.currentTimeMillis()` which may or may not be UTC, depending on the server's settings. This can lead to incorrect metrics being reported by Split in the portal. Additionally, if there's some problem in our pipeline, there's no way to replay events and have them show up at the proper time. This ends up mattering a lot sometimes. So, instead of re-implementing all our logic to use the HTTP Admin API, it made a lot more sense to just open a pull request against the Java SDK.

This PR adds support for timestamps to the Split Client's `.track(...)` functions. It also updates the test suite in order to assert the behavior and increments the client's patch version (though perhaps it should have been a minor version increment instead?).

### DIFF
--- a/client/CHANGES.txt
+++ b/client/CHANGES.txt
@@ -1,5 +1,8 @@
 CHANGES
 
+3.2.3
+- Add support for optional timestamp parameter when tracking an event.
+
 3.2.2
 - log warn and not error when Split doesn't exist in the environment
 
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.split.client</groupId>
         <artifactId>java-client-parent</artifactId>
-        <version>3.2.2</version>
+        <version>3.2.3</version>
     </parent>
     <artifactId>java-client</artifactId>
     <packaging>jar</packaging>
--- a/client/src/main/java/io/split/client/AlwaysReturnControlSplitClient.java
+++ b/client/src/main/java/io/split/client/AlwaysReturnControlSplitClient.java
@@ -68,7 +68,27 @@ public class AlwaysReturnControlSplitClient implements SplitClient {
     }
 
     @Override
+    public boolean track(String key, String trafficType, String eventType, long timestamp) {
+        return false;
+    }
+
+    @Override
     public boolean track(String key, String trafficType, String eventType, double value, Map<String, Object> properties) {
+        return false;
+    }
+
+    @Override
+    public boolean track(String key, String trafficType, String eventType, double value, long timestamp) {
+        return false;
+    }
+
+    @Override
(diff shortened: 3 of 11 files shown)


patricioe:
Hey,

Thank for reaching out. The solution is not as simple as it seems.

Time consistency is an important factor at Split. Our (patented) attribution algorithm (the ones that finds all events that fall for a given user and treatment) and highly dependent to the time reported in the impressions (this is when evaluating flags). If we let customers add their own time, aside from the EPOC time reported by the System, it's create an interesting complication for that to the point that our support team will never be able to figure out.

I'm curious why you think Sytem.currentTimeInMillis() is not correct?Mind elaborating the issue you are facing and for which you need this change?

https://stackoverflow.com/questions/17271039/does-system-currenttimemillis-return-utc-time

`Your local time zone doesn't affect how many milliseconds have occurred since that epoch`

valesken:
@patricioe thank you for your response! Turns out I was mixing up the implementations of `System.currentTimeInMillis()` and `new Date(System.currentTimeInMillis())`, the latter of which does invoke the local timezone.

That said, I'm still not quite sure I understand why this is a problematic change, especially since the capability already exists in your RESTful api: https://docs.split.io/reference#create-event. I just don't want to have to reimplement our code to use the api instead of your SDK. But honestly, without this change, I'll have to.

The issue we're facing is that, due to one bug or another on our side, we're unable to send some events to Split. For example, an event could come appear in our queue and we fail to parse it properly and so determine that we shouldn't send it to split. Once we fix the bug, events flow properly again. But with this SDK, there's no way to replay the events that we missed and have them show up at the proper time. Occasionally, this completely throws off our experiments. There's 3 things we can do:
1) Replay the events and have them appear as a giant spike of events that all happen at the same point in time. This is obviously incorrect in terms of determining the result of an experiment. Especially if we replay after the experiment has concluded.
2) Do not replay the events. This is also obviously incorrect because it may look like some treatment actually caused a reduction in certain kinds of events (when in fact, it was just a separate, unrelated bug).
3) Replay the events with the correct timestamp. This makes the statical results of our experiments correct, but then we can't use the SDK because it doesn't support us adding timestamps.

And there's another issue. Say an event occurs in some offline client and, once it comes online, the event is sent to our server. If we try to process it, the timestamp that is applied will again be simply wrong.

I get that you're worried about customers unwittingly sending events that "don't show up" because they set the wrong timestamp. But the trade off is that you're making it so that customers can't recover their experiments if anything goes wrong.

patricioe:
@valesken 
Thanks for clarifying further. I understand where you are coming from.

Let me take it for another spin internally and get back to you.

valesken:
Hi @patricioe have you had another chance to look at this issue? It's come up again and we'd very much appreciate support on this if possible.

PR #96, opened 2019-08-15
Switch to Ubuntu Trusty

Israphel:
Using JDK 1.7 implies Maven2 and the Sonar plugin is deprecated (it won't work with our Sonar server).

Since releases are being made with JDK 1.8 anyway, let's run the tests with the same version.


splitadricejas:
LGTM

PR #124, opened 2020-06-26
Impression seenAt

mredolatti:


client/src/main/java/io/split/client/impressions/ImpressionObserver.java
```diff
@@ -0,0 +1,38 @@
...
+Since the library is shaded and should not be exposed to users of the SDK, it's safe to use it here.
+ */
+
+@SuppressWarnings("UnstableApiUsage")
+public class ImpressionObserver {
+
+    private final Cache<Long, Long> _cache;
+
+    public ImpressionObserver(long size) {
+        _cache = CacheBuilder.newBuilder()
```
patricioe, line 20: do we need to set the concurrency level?

patricioe: how about the expiration settings?

client/src/main/java/io/split/client/impressions/ImpressionObserver.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public Long testAndSet(KeyImpression impression) {
+        if (null == impression) {
+            return null;
+        }
+
+        Long hash = ImpressionHasher.process(impression);
+
+        synchronized(this) {
```
patricioe, line 32: this can't be here. It'll be too slow. Better to use the compute method from guava cache.

patricioe: I guess the caller is single thread correct?

client/src/main/java/io/split/client/impressions/ImpressionObserver.java
mredolatti: Regarding cuncurrency level: the cache is only used from the impressions manager which is executed within a singleThreadedExecutor, so we should not have race conditions here, nor benefit from partitioning to minimize contention. I can however set the default value (4) explicitly, in case it changes sometime, or we decide to go multithreaded here

Regarding expiration settings: We discussed this thoroughly with nico, and in order to minimize memory usage, we decided not to attach an extra TTL to each cache item, since the difference between sending `null` or sending an old timestamp doesn't really make a difference. We let the events server decide whether the timestamp is recent enough to filter the impression from the exp pipeline or not

client/src/main/java/io/split/client/impressions/ImpressionObserver.java
mredolatti: I'm not familiar with `compute` method, i'll look into it.
It's single threaded, but I added the synchronized block, in order to avoid surprises in the future, are those statements really that slow? the hashing computation is done outside of the critical region

PR #166, opened 2020-11-06
use milliseconds instead of seconds

mredolatti:


client/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>io.split.client</groupId>
         <artifactId>java-client-parent</artifactId>
-        <version>4.1.2</version>
+        <version>4.1.3</version>
```
sanzmauro, line 8: The version should be 4.1.2.

sanzmauro (approved):
lgtm

mredolatti:
Tackled in other PR

PR #189, opened 2021-02-24
Turning off logging when refresh is forced.

ldecheverz-split:


ldecheverz-split:
Including this feature in other PR

PR #192, opened 2021-03-10
Connection confirmation first event

ldecheverz-split:
Connection First Event
Fixes:
-Log on reset
-Backoff
-Loop after refresh
-Typo


sanzmauro (approved):
lgtm

PR #206, opened 2021-04-16
Do not embed JUnit / Set scope to provided on testing client

gb:
[Snyk](https://snyk.io/) is now present in the build process at Salesforce projects.
While checking the vulnerabilities reports, I notice an older version of JUnit coming from the Split testing jar.
Yes, we were using an outdated version of the Split SDK, and upgrading it to the latest fixed it. 

But to avoid this kind of issue in the future, I would like to suggest this change.
IMO, JUnit should be provided by the consumer, not by the library, and will likely be already present in the project.

Thank you!

### DIFF
--- a/testing/pom.xml
+++ b/testing/pom.xml
@@ -13,6 +13,7 @@
     <packaging>jar</packaging>
     <name>Java Client For Testing</name>
     <description>Testing suite for Java SDK for Split</description>
+
     <dependencies>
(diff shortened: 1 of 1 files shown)


nmayorsplit:
Hi @gb 
We released Java SDK 4.4.6 version that contains this update.

Thanks
Nadia

PR #207, opened 2021-04-16
Readme & pom fix

dogeared:
If you attempt to build with an earlier jdk, you get:

```
java: warning: source release 8 requires target release 1.8
```

because of the source and target definitions in the pom.xml.

This also updates pom.xml's to use the properties defined in the parent for source and target


nmayorsplit:
Hi @dogeared 
We released Java SDK 4.4.6 version that contains these updates.

Thanks
Nadia

PR #234, opened 2021-06-08
NOC-608: Adding extra Host header for Fastly

gpochettino:
Adding extra Host header for Segment & Split Changes. In that way we can set the SDK to directly call Fastly DCA.

```java
SplitClientConfig config = SplitClientConfig.builder()
                .endpoint("https://cache-dca17775.hosts.fastly.net", "https://events.split-stage.io")
                .setBlockUntilReadyTimeout(60000)
                .hostHeader("sdk.split-stage.io")
                .build();
```


NicoZelaya:
I'll close the PR without touching the branch for now.

PR #247, opened 2021-07-13
Add Support for Cucumber using @split[feature:treatment] tags

aslakhellesoy:
Hi Splitsters,

This PR adds a small utility that would make it easier to use Split from [Cucumber](https://cucumber.io/) Scenarios. Here is a short example:

```gherkin
Feature: Make Coffee

  @split[cappuccino:on]
  Scenario: verify functionality with cappuccino feature
    Given ...
    When ...
    Then ...

  @split[cappuccino:off]
  Scenario: verify functionality without cappuccino feature
    Given ...
    When ...
    Then ...
```

In order to enable the integration with Split, a [Before Hook](https://cucumber.io/docs/cucumber/api/#hooks) must be defined:

```java
import io.cucumber.java.Before;
import io.split.client.testing.SplitClientForTest;
import io.split.client.testing.cucumber.CucumberSplit;

public class StepDefinitions {
    private final SplitClientForTest splitClient = new SplitClientForTest();

    @Before
    public void configureSplit(Scenario scenario) {
        CucumberSplit.configureSplit(splitClient, scenario);
    }
}
```

I hope you'll consider adding this to the `io.split.client:java-client-testing` library!

Many thanks,

Aslak Hellesøy
Creator and lead developer of Cucumber Open


testing/pom.xml
```diff
@@ -23,5 +23,15 @@
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>cucumber-java</artifactId>
+            <version>6.10.4</version>
```
aslakhellesoy, line 29: This should probably be test scope, no?

patricioe: probably

testing/src/test/resources/io/split/client/testing/cucumber/Split.feature
```diff
@@ -0,0 +1,19 @@
+@split[cappuccino:off]
+@split[dollars:on]
+Feature: Split
```
patricioe, line 3: Would you mind educating me how the unit test binds to the split listed in this file?

also, is the idea to have a single descritptor file or one per test class or test case?

testing/src/main/java/io/split/client/testing/cucumber/CucumberSplit.java
```diff
@@ -0,0 +1,51 @@
...
+ * }
+ * </pre>
+ */
+public class CucumberSplit {
+    private static final Pattern SPLIT_TAG_PATTERN = Pattern.compile("^@split\\[(.*):(.*)]");
+
+    public static void configureSplit(SplitClientForTest splitClient, Scenario scenario) {
+        Collection<String> tags = scenario.getSourceTagNames();
+        for (String tag : tags) {
+            Matcher matcher = SPLIT_TAG_PATTERN.matcher(tag);
```
patricioe, line 43: Question here:
- if more than 1 tag matches this, do you end up overwriting the registered (feature, treatment) pair?

testing/pom.xml
aslakhellesoy: I've thought a bit more about this. I think there are two options here - `compile` scope or `provided` scope. (See [maven docs](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope) for details).

The `junit` dependency is already `compile` scope, so using the same scope for `cucumber-java` and `cucumber-junit` would be consistent with that.

Alternatively, making it `provided` scope would indicate that dependent modules would have to explicitly add the dependencies themselves. I think that's ok too.

I'll defer to you to decide what's most appropriate.

testing/src/test/resources/io/split/client/testing/cucumber/Split.feature
aslakhellesoy: I'm not sure what you mean by unit test or descriptor file. I'll clarify some of the terminology used by Cucumber (which is closer to an acceptance testing tool than a unit testing tool).

A `.feature` file (sometimes called *gherkin document*) contains multiple *scenarios* (sometimes called *examples*).
Each *Scenario* is an acceptance test and consists of the following parts:

* One or more [Steps](https://cucumber.io/docs/gherkin/reference/#example) (`Given`, `When`, `Then`).
* [Step Definitions](https://cucumber.io/docs/cucumber/step-definitions/), which are Java methods annotated with `@Given`, `When` or `@Then`.

Cucumber creates a new instance of every class with step definitions before the scenario runs. This is similar to how JUnit creates a new instance of a test class for every `@Test` method it executes.

For each step in the scenario, Cucumber then finds a matching step definition and invokes the method.

Step definition classes aren't scoped to a particular `.feature` file. The steps in a `.feature` file can invoke step definitions from any step definition class - the matching is based on the expression in the step definition annotation.

Scenarios can have [tags](https://cucumber.io/docs/cucumber/api/?sbsearch=tag#tags). The code in this library looks for tags of the format `@split[feature:treatment]` and uses that to register the appropriate treatments.

testing/src/main/java/io/split/client/testing/cucumber/CucumberSplit.java
aslakhellesoy: Yes. This is intentional.

Tags that are placed at the top level `Feature` level are inherited by all the `Scenario`s. The `getSourceTagNames()` method returns a `Collection` of the tags from both the `Feature` and the current `Scenario`. The inherited `Feature` tags are always before the `Scenario` tags in the collection (it would have been more appropriate to declare the return type as `List` rather than `Collection` since it's ordered).

We're taking advantage of this here to allow users to define "default" treatments at the feature level, and (optionally) override them at the scenario level. I have illustrated this in the last scenario where the default `dollars=off` treatment from the feature level is overridden to `dollars=on` for that particular scenario.

aslakhellesoy: I've pushed some more changes to (hopefully) illustrate better how this works.

testing/pom.xml
patricioe: If someone is already using this framework (our Java-testing) without cucumber and I merge this PR. Regardless of the provided or not, I believe the build will fail for not finding the dependency in the classpath, right? 

I need time to sync internally but if that's the case we may need to split this into a separate submodule.

patricioe:
@aslakhellesoy The only dependency with Split testing module is on the SplitClientForTest correct? just triple checking.

testing/pom.xml
aslakhellesoy: > I believe the build will fail for not finding the dependency in the classpath, right?

No it won't.

In Java, classes are only loaded when they are referenced by another class. Classes are not loaded by mere presence in a jar file.

No other classes in this library reference `io.split.client.testing.cucumber.CucumberSplit`, so it won't be loaded.

aslakhellesoy:
> @aslakhellesoy The only dependency with Split testing module is on the SplitClientForTest correct? just triple checking.

That is correct.

The only class I have added to the module is `io.split.client.testing.cucumber.CucumberSplit`. As you can see from its `import` statements, its dependencies are:

```java
import io.cucumber.java.Scenario;
import io.split.client.testing.SplitClientForTest;

import java.util.Collection;
import java.util.regex.Matcher;
import java.util.regex.Pattern;
```

patricioe:
hey @aslakhellesoy are you able to allow me to edit your fork pull request? I'd like to create a new module so you have complete isolation for iterations and so I can run a git checkout on this branch to do a release.

image

patricioe:
Let's continue in https://github.com/splitio/java-client/pull/248

aslakhellesoy:
I've now moved this code to https://github.com/cucumber/split-java/ and added some documentation. Please remove the code I have contributed now that Cucumber-Split has a new home.

PR #248, opened 2021-07-26
Cucumber

patricioe:
This is a refactor from pull request https://github.com/splitio/java-client/pull/247 which incorporates the changes into a separate maven module for ease of integration.


patricioe:
@aslakhellesoy could you pease give it a try before a deploy to maven this artifact?

aslakhellesoy:
Thanks @patricioe - I'll take it for a spin....

testing-cucumber/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+            <version>${parent.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>cucumber-java</artifactId>
+            <version>6.10.4</version>
```
aslakhellesoy, line 30: I would add `<scope>provided</scope>` here so that users can choose a more recent version of Cucumber when it's released. (The API we're using is stable and unlikely to change).

testing-cucumber/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.cucumber</groupId>
+            <artifactId>cucumber-junit</artifactId>
+            <version>6.10.4</version>
```
aslakhellesoy, line 35: I would add `<scope>test</scope>` here since it's not used in `src/main`, only in `src/test`.

aslakhellesoy (changes requested):
I've suggested two changes to `pom.xml`.

I also suggest adding this to the root `pom.xml` so the module gets included in the root level build:

```xml
    <modules>
        <module>testing</module>
        <!-- Add this line -->
        <module>testing-cucumber</module>
        <module>client</module>
    </modules>
```

Thanks a lot!!

patricioe:
@aslakhellesoy done

patricioe:
@aslakhellesoy deployed. java-client-testing-cucumber-1.0.0-rc1.jar

I won't merge the PR yet.

aslakhellesoy:
I've now moved this code to https://github.com/cucumber/split-java/ and added some documentation. Please remove the code I have contributed now that Cucumber-Split has a new home.

PR #278, opened 2021-12-10
Get treatments

ldecheverz-split:


mredolatti (approved):

stil4m:
@ldecheverz-split This change removed the `AlwaysReturnControlSplitClient.java`. What is the expected upgrade path if you are using this?

ldecheverz-split:
Hi Mats!
First of all, sorry for the inconvenience. This class should not be part of our Public API. That's the reason we removed it. In case you are using this class for testing, we have a module for it where you can use a class named SpliClientForTest. This class allows you to also add particular treatments result if you want or by default returning CONTROL:

    

    SplitClientForTest splitClientForTest= new SplitClientForTest(); 
    String controlTreatment = splitClientForTest.getTreatment("myKey", "mySplit");


Another option is to mock interactions with the SDK via Mockito: 


        MyApp myApp = new MyApp(Mockito.mock(SplitClient.class));

        Mockito.when(client.getTreatment(Mockito.anyString(), Mockito.anyString())).thenReturn("CONTROL");

        //Calling method that checks treatments

        myApp.interactWithSplit();

If none of these options work for your use case we would be interested in diving a bit deeper into it. If this is the case it would be great if you can contact us via support@split.io to discuss and potentially book a call.
Cheers

PR #295, opened 2022-04-28
Fix not relocated packages

maciejnawrocki:
Hi, 

I added relocations of transitive dependencies shadowed in java-client.jar, because my team has a problem with duplicated classes for:
`org.apache.httpcomponents.client5:httpclient5`

I also added other relocations for `org.checkerframework` and `org.yaml.snakeyaml` to avoid potential further problems

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -61,8 +61,16 @@
                             </transformers>
                             <relocations>
                                 <relocation>
-                                    <pattern>org.apache.http</pattern>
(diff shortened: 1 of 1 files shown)


mmelograno:
Hi @maciejnawrocki,

Thanks for providing the snippet for shading dependencies. The PR was merged in our latest release.

I'll proceed to Close this PR.

Cheers,

PR #305, opened 2022-05-30
BloomFilter Implementation

nmayorsplit:
I created the CuckooFilter and FilterAdapter.
Instead of using BloomFilter, I used CuckoFilter because the last one has the clean method.


mredolatti (approved):
LGTM!

PR #306, opened 2022-06-03
UniqueKeysTracker Implementation

nmayorsplit:


client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+public class UniqueKeysTrackerImp implements UniqueKeysTracker{
+    private static final double MARGIN_ERROR = 0.01;
+    private static final int MAX_AMOUNT_OF_TRACKED_MTKS = 30000;
+    private static final int MAX_AMOUNT_OF_KEYS = 10000000;
+    private FilterAdapter filterAdapter;
+    private ImpressionsSender impressionsSender;
+    private final ConcurrentHashMap<String,HashSet<String>> mtkTracker;
+    private static final Logger _logger = LoggerFactory.getLogger(HttpImpressionsSender.class);
```
mmelograno, line 20: the name should be `UniqueKeysTrackerImp.class` instead of `HttpImpressionSender.class`

client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public boolean track(String featureName, String key) {
+        if (mtkTracker.size() == MAX_AMOUNT_OF_TRACKED_MTKS){
+            //@todo implements add logic to flush data when the Dictionary is complete.
+            //flush
+            _logger.warn("The MTKTracker size reached the maximum limit");
+            return false;
+        }
+        if (mtkTracker.size() < MAX_AMOUNT_OF_TRACKED_MTKS) {
```
mmelograno, line 36: the if is unnecessary due to the previous return

client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java
```diff
@@ -0,0 +1,64 @@
...
+            if (filterAdapter.add(featureName, key)) {
+                HashSet<String> value = new HashSet<>();
+                if(mtkTracker.containsKey(featureName)){
+                    value = mtkTracker.get(featureName);
+                }
+                value.add(key);
+                mtkTracker.put(featureName, value);
+                return true;
+            }
+            _logger.warn("The feature and key pair was added");
```
mmelograno, line 46: the log should be debug because is not a waning situation, and you can log the specific key par too

client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public void start() {
+
+    }
+
+    @Override
+    public void stop() {
+
+    }
+    ConcurrentHashMap<String, HashSet<String>> getMtkTracker() {
```
mmelograno, line 61: maybe this method should be visible only for testing

mmelograno (changes requested):

client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java
nmayorsplit: I added the annotation VisibleForTesting.

client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java
nmayorsplit: I changed it.

client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java
nmayorsplit: I removed it.

mmelograno (approved):
LGTM!

PR #307, opened 2022-06-07
[SDKS-5852] Add logic for NONE mode in ImpressionManager

nmayorsplit:
- Create ProcessImprerssionStrategy
- Add UniqueKeysTracker in the ImpressionManager


client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionMTK.java
```diff
@@ -0,0 +1,25 @@
+package io.split.client.impressions.strategy;
+
+import io.split.client.impressions.Impression;
+import io.split.client.impressions.ImpressionsResult;
+import io.split.client.impressions.UniqueKeysTracker;
+
+import java.util.List;
+
+public class ProcessImpressionMTK implements ProcessImpressionStrategy{
```
sanzmauro, line 9: Please rename this to `ProcessImpressionNone`. 

We have 3 strategies: `debug`, `optimized` and `none` so this request is to keep the naming convention.

client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionMTK.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public ProcessImpressionMTK(UniqueKeysTracker uniqueKeysTracker) {
+        _uniqueKeysTracker = uniqueKeysTracker;
+    }
+
+    @Override
+    public ImpressionsResult processImpressions(List<Impression> impressions) {
+
+        for(Impression impression: impressions){
+            _uniqueKeysTracker.track(impression.split(),impression.key());
```
sanzmauro, line 21: When the sdk is in `none` mode, we also should to keep the total amount of evaluations. 
You should to add the impression count here as you are using in optimized mode.

client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionStrategy.java
```diff
@@ -0,0 +1,11 @@
+package io.split.client.impressions.strategy;
+
+import io.split.client.impressions.Impression;
+import io.split.client.impressions.ImpressionsResult;
+
+import java.util.List;
+
+public interface ProcessImpressionStrategy {
+
+    ImpressionsResult processImpressions(List<Impression> impressions);
```
sanzmauro, line 10: maybe the name can be just `process` ??? because the interface name is already `processImpression...`

client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionOptimized.java
```diff
@@ -0,0 +1,38 @@
...
+        _impressionCounter = impressionCounter;
+    }
+
+    @Override
+    public ImpressionsResult processImpressions(List<Impression> impressions) {
+        List<Impression> impressionsToQueue = new ArrayList<>();
+        for(Impression impression : impressions) {
+            impression = impression.withPreviousTime(_impressionObserver.testAndSet(impression));
+            _impressionCounter.inc(impression.split(), impression.time(), 1);
+            if(!shouldQueueImpression(impression)) {
```
sanzmauro, line 26: Why you are using a negative if?
Personally in this case I prefer to use a positive if and avoid the `continue` statement.

client/src/main/java/io/split/client/impressions/ImpressionsManagerImpl.java
```diff
@@ -96,17 +95,21 @@ private ImpressionsManagerImpl(CloseableHttpClient client,
                 : new ImpressionListener.NoopImpressionListener();
 
         _operationMode = config.operationMode();
-        _addPreviousTimeEnabled = shouldAddPreviousTime();
-        _counter = _addPreviousTimeEnabled ? new ImpressionCounter() : null;
-        _isOptimized = _counter != null && shouldBeOptimized();
-        if (_isOptimized) {
+        ProcessImpressionFactory processImpressionFactory =  new ProcessImpressionFactory();
+        if (_config.impressionsMode().equals(Mode.OPTIMIZED)){
```
sanzmauro, line 99: you are asking twice for the impression mode, we should chat this in private to improve it.

sanzmauro (changes requested):
I think that you can add more tests for this code, eg for each strategy.

client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionMTK.java
nmayorsplit: I changed the name.

client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionMTK.java
nmayorsplit: I added the impression count.

client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionOptimized.java
nmayorsplit: I removed the not and changed the logic in the method.

client/src/main/java/io/split/client/impressions/ImpressionsManagerImpl.java
nmayorsplit: I removed the Factory class. Now we are checking the mode in the ImpressionManager class.

mmelograno (approved):

PR #313, opened 2022-07-11
Task/pipeline

mmelograno:
- Implement pipeline for Pluggable and Redis.


redis-wrapper/src/main/java/redis/RedisPipeline.java
```diff
@@ -0,0 +1,63 @@
...
+    private final JedisPool _jedisPool;
+
+
+    public RedisPipeline(JedisPool jedisPool, String prefix) {
+        _jedisPool = jedisPool;
+        _prefix = prefix;
+        try (Jedis jedis = jedisPool.getResource()) {
+            _pipelined = jedis.pipelined();
+        } catch (Exception ex) {
+            System.out.println("err " + ex.getMessage());
```
mmelograno, line 23: it should throws RedisException

redis-wrapper/src/main/java/redis/RedisPipeline.java
```diff
@@ -0,0 +1,63 @@
...
+    public void delete(List<String> keys){
+        if(keys == null || keys.isEmpty()){
+            return ;
+        }
+        try (Jedis jedis = _jedisPool.getResource()) {
+            keys = keys.stream().map(key -> buildKeyWithPrefix(key)).collect(Collectors.toList());
+
+            jedis.del(keys.toArray(new String[keys.size()]));
+        } catch (Exception ex) {
+            throw ex;
```
mmelograno, line 50: it should throws RedisException

nmayorsplit: Changed it.

redis-wrapper/src/main/java/redis/RedisPipeline.java
nmayorsplit: Changed it.

redis-wrapper/src/main/java/redis/RedisPipeline.java
```diff
@@ -0,0 +1,65 @@
...
+        _jedisPool = jedisPool;
+        _prefix = prefix;
+        try (Jedis jedis = _jedisPool.getResource()) {
+            _pipelined = jedis.pipelined();
+        } catch (Exception ex) {
+            new RedisException(ex.getMessage());
+        }
+    }
+
+    /* package private */ String buildKeyWithPrefix(String key) {
```
mmelograno, line 30: this method is used in several places, can we refactor to have in some common place?

redis-wrapper/src/main/java/redis/RedisPipeline.java
```diff
@@ -0,0 +1,65 @@
...
+        if (!key.startsWith(_prefix)) {
```
mmelograno, line 31: this checking is an extra thing that should be removed, if for some reason two prefix are concatenated, definitely we have a bug in the code

redis-wrapper/src/main/java/redis/RedisPipeline.java
nmayorsplit: I'm going to work on this using another branch.

redis-wrapper/src/main/java/redis/RedisPipeline.java
nmayorsplit: The same before comment.

client/src/main/java/io/split/storages/pluggable/domain/UserPipelineWrapper.java
```diff
@@ -0,0 +1,39 @@
...
+        _pipeline = pipeline;
+    }
+
+    @Override
+    public List<Result> exec() throws Exception {
+        try{
+            return _pipeline.exec();
+        } catch (Exception e) {
+            _logger.warn("Exception calling Pipeline exec", e);
+            throw new Exception(e.getMessage());
```
mmelograno, line 27: you can return an empty list

nmayorsplit: I added the return.

PR #335, opened 2022-08-31
[SDKS-5975] Issues and PRs in Github

nmayorsplit:
Some suggestions from PR and issues in GitHub:

- Update pom files 
- Move change logs to the project root
- Update the Java version in the readme.md


pom.xml
```diff
@@ -93,8 +93,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.3</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>${maven.compiler.source}</source>
+                    <target>${maven.compiler.target}</target>
```
gthea, line 97: This value would be 8 instead of 1.8, I guess that's fine?

gthea (approved):

pom.xml
nmayorsplit: It's fine to tell Maven that uses Java 8 adding source and target 1.8.

PR #337, opened 2022-09-06
[SDKS-6101] Move UniqueKeysTracker start to SyncManager and SplitFactory

nmayorsplit:
After merging the destroy refactor, I moved the start UniqueKeys to SyncManager and created UniqueKeysTracker in SplitFactory.

### DIFF
--- a/client/src/main/java/io/split/client/SplitFactoryImpl.java
+++ b/client/src/main/java/io/split/client/SplitFactoryImpl.java
@@ -137,7 +137,7 @@ public class SplitFactoryImpl implements SplitFactory {
     private final ImpressionsSender _impressionsSender;
     private final URI _rootTarget;
     private final URI _eventsRootTarget;
-    private UniqueKeysTracker uniqueKeysTracker;
+    private final UniqueKeysTracker _uniqueKeysTracker;
 
 
     //Constructor for standalone mode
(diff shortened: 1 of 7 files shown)


client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -280,8 +291,22 @@ protected SplitFactoryImpl(String apiToken, SplitClientConfig config, CustomStor
 
         _evaluator = new EvaluatorImp(userCustomSplitAdapterConsumer, userCustomSegmentAdapterConsumer);
         _impressionsSender = PluggableImpressionSender.create(customStorageWrapper);
+
+        if (config.impressionsMode().equals(ImpressionsManager.Mode.NONE)){
+            int uniqueKeysRefreshRate = config.operationMode().equals(OperationMode.STANDALONE) ? config.uniqueKeysRefreshRateInMemory()
+                    : config.uniqueKeysRefreshRateRedis();
+            uniqueKeysTracker = new UniqueKeysTrackerImp(_telemetrySynchronizer, uniqueKeysRefreshRate, config.filterUniqueKeysRefreshRate());
+        }
```
gthea, line 299: Maybe this could be extracted to a method? since it's being used in both constructors.
```suggestion
        uniqueKeysTracker = buildUniqueKeysTracker(_telemetrySynchronizer, config);
```

client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -134,6 +137,7 @@ public class SplitFactoryImpl implements SplitFactory {
     private final ImpressionsSender _impressionsSender;
     private final URI _rootTarget;
     private final URI _eventsRootTarget;
+    private UniqueKeysTracker uniqueKeysTracker;
```
gthea, line 140: I think this one can be final
```suggestion
    private final UniqueKeysTracker uniqueKeysTracker;
```

client/src/main/java/io/split/client/SplitFactoryImpl.java
nmayorsplit: I created a method for this. Thank you!

client/src/main/java/io/split/client/SplitFactoryImpl.java
nmayorsplit: I made it final.

gthea (approved):

PR #357, opened 2022-11-28
[SDKS-6274] Create Interface SplitSyncTask

nmayorsplit:


client/src/main/java/io/split/engine/experiments/SplitSyncTask.java
```diff
@@ -0,0 +1,8 @@
+package io.split.engine.experiments;
+
+public interface SplitSyncTask {
+
+    void startPeriodicFetching();
```
mmelograno, line 5: let's call it start

mmelograno (changes requested):

client/src/main/java/io/split/engine/experiments/SplitSyncTask.java
nmayorsplit: I changed it.

sanzmauro (approved):

mmelograno (approved):

PR #358, opened 2022-11-30
 [SDKS-6270] Create LocalhostSplitFetcher

nmayorsplit:


client/src/main/java/io/split/engine/experiments/LocalhostSplitChangeFetcher.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Override
+    public SplitChange fetch(long since, FetchOptions options) {
+        SplitChange splitChange = null;
+        try {
+            JsonReader jsonReader = new JsonReader(new FileReader(_file));
+
+            Gson gson = new Gson();
+            splitChange = gson.fromJson(jsonReader, SplitChange.class);
```
gthea, line 29: If it's worth it, maybe this `fromJson(JsonReader jsonReader, Class<T> class)` method could be added to the `io.split.client.utils.Json` class to avoid creating the Gson instance.
```suggestion
            splitChange = Json.fromJson(jsonReader, SplitChange.class);
```

client/src/main/java/io/split/engine/experiments/LocalhostSplitChangeFetcher.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.io.File;
+import java.io.FileReader;
+
+public class LocalhostSplitChangeFetcher implements SplitChangeFetcher {
+
+    private static final Logger _log = LoggerFactory.getLogger(LocalhostSplitChangeFetcher.class);
+    private final File _file;
+
+    public LocalhostSplitChangeFetcher(String directory, String fileName){
```
mmelograno, line 18: let's use full path

client/src/main/java/io/split/engine/experiments/LocalhostSplitChangeFetcher.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public SplitChange fetch(long since, FetchOptions options) {
+        SplitChange splitChange = null;
+        try {
+            JsonReader jsonReader = new JsonReader(new FileReader(_file));
+
+            Gson gson = new Gson();
+            splitChange = gson.fromJson(jsonReader, SplitChange.class);
+        } catch (Exception e) {
+            _log.warn("There was no file named " + _file.getPath() + " found. " +
```
mmelograno, line 33: maybe using String.format here instead of concatenating

client/src/test/resources/split_init.json
```diff
@@ -0,0 +1,567 @@
+{
+  "splits": [
+    {
+      "trafficTypeName": "user",
+      "name": "bilal_split",
```
mmelograno, line 5: let's use names just like split_1 ...

mmelograno (changes requested):

client/src/test/java/io/split/engine/experiments/LocalhostSplitChangeFetcherTest.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Test
+    public void testParseSplitChange(){
+        LocalhostSplitChangeFetcher localhostSplitChangeFetcher = new LocalhostSplitChangeFetcher("src/test/resources", "split_init.json");
+        FetchOptions fetchOptions = Mockito.mock(FetchOptions.class);
+
+        SplitChange splitChange = localhostSplitChangeFetcher.fetch(-1L, fetchOptions);
+
+        List<Split> split = splitChange.splits;
+        Assert.assertEquals(7, split.size());
```
gthea, line 22: As an easy verification maybe `till` and `since` can be checked as well?

client/src/main/java/io/split/engine/experiments/LocalhostSplitChangeFetcher.java
nmayorsplit: Added.

client/src/main/java/io/split/engine/experiments/LocalhostSplitChangeFetcher.java
nmayorsplit: Added.

client/src/test/resources/split_init.json
nmayorsplit: Changed split names and segments.

client/src/main/java/io/split/engine/experiments/LocalhostSplitChangeFetcher.java
nmayorsplit: Added new method.

client/src/test/java/io/split/engine/experiments/LocalhostSplitChangeFetcherTest.java
nmayorsplit: Added.

mmelograno (approved):

PR #365, opened 2022-12-21
fix: updating logs to use '{}' instead of '%s'

ClaudenirFreitas:


nmayorsplit:
I'm going to close it because the 4.7.0 version fixed it.

PR #368, opened 2023-01-04
[SDKS-6387] Add SegmentSynchronizationTask to current flow

nmayorsplit:
- Add Segment SegmentSynchronizationTask to the current flow
- Add segmentDirectory to the config
- Fix log message in SplitClientImpl

### DIFF
--- a/client/src/main/java/io/split/client/SplitFactoryBuilder.java
+++ b/client/src/main/java/io/split/client/SplitFactoryBuilder.java
@@ -39,6 +39,9 @@ public class SplitFactoryBuilder {
     public static synchronized SplitFactory build(String apiToken, SplitClientConfig config) throws IOException, URISyntaxException {
         ApiKeyValidator.validate(apiToken);
         if (LocalhostSplitFactory.LOCALHOST.equals(apiToken)) {
+            if (config.splitFile().endsWith("json")){
+                return new SplitFactoryImpl(config);
+            }
             return LocalhostSplitFactory.createLocalhostSplitFactory(config);
         }
         if (StorageMode.PLUGGABLE.equals(config.storageMode()) || StorageMode.REDIS.equals(config.storageMode())){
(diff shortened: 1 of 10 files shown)


client/src/main/java/io/split/client/SplitFactoryBuilder.java
```diff
@@ -39,6 +39,9 @@ public static SplitFactory build(String apiToken) throws IOException, URISyntaxE
     public static synchronized SplitFactory build(String apiToken, SplitClientConfig config) throws IOException, URISyntaxException {
         ApiKeyValidator.validate(apiToken);
         if (LocalhostSplitFactory.LOCALHOST.equals(apiToken)) {
+            if (config.splitFile().endsWith("json")){
```
mmelograno, line 42: I think that you should cover `.json` and also put everything in lowercase just in case

client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -322,6 +327,94 @@ protected SplitFactoryImpl(String apiToken, SplitClientConfig config, CustomStor
         manageSdkReady(config);
     }
 
+    // Localhost
+    protected SplitFactoryImpl(SplitClientConfig config) throws URISyntaxException {
+        _userStorageWrapper = null;
+        _operationMode = config.operationMode();
+        _startTime = System.currentTimeMillis();
```
mmelograno, line 334: not needed

client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -322,6 +327,94 @@ protected SplitFactoryImpl(String apiToken, SplitClientConfig config, CustomStor
         manageSdkReady(config);
     }
 
+    // Localhost
+    protected SplitFactoryImpl(SplitClientConfig config) throws URISyntaxException {
+        _userStorageWrapper = null;
+        _operationMode = config.operationMode();
```
mmelograno, line 333: not needed

mmelograno (changes requested):

sanzmauro (approved):

client/src/main/java/io/split/client/SplitFactoryImpl.java
nmayorsplit: I removed it and removed final word.

client/src/main/java/io/split/client/SplitFactoryImpl.java
nmayorsplit: I removed it and removed final word.

client/src/main/java/io/split/client/SplitFactoryBuilder.java
nmayorsplit: Now we check that the file could be json, JSON or Json.

mmelograno (approved):

PR #369, opened 2023-01-05
[SDKS-6350] Add refreshEnable config for localhost

nmayorsplit:
- Add localhostRefreshEnable config
- Fix the case that the segmentDirectory config is null.


mredolatti (approved):
lgtm

sanzmauro (approved):

PR #375, opened 2023-01-27
[SDKS-6389] Add localhost JSON sanitization

nmayorsplit:


client/pom.xml
```diff
@@ -177,6 +177,11 @@
             <artifactId>snakeyaml</artifactId>
             <version>1.32</version>
         </dependency>
+        <dependency>
```
mmelograno, line 180: why we need a new dependency for parsing JSON?

client/src/main/java/io/split/client/LocalhostSegmentChangeFetcher.java
```diff
@@ -24,8 +24,9 @@ public LocalhostSegmentChangeFetcher(String filePath){
     @Override
     public SegmentChange fetch(String segmentName, long changesSinceThisChangeNumber, FetchOptions options) {
         try {
-            JsonReader jsonReader = new JsonReader(new FileReader(String.format("%s/%s.json", _file, segmentName)));
-            return Json.fromJson(jsonReader, SegmentChange.class);
+            ObjectMapper objectMapper = new ObjectMapper();
```
mmelograno, line 27: JsonReader jsonReader = new JsonReader(new FileReader(String.format("%s/%s.json", _file, segmentName)));
            SegmentChange segmentChange = Json.fromJson(jsonReader, SegmentChange.class);
            return LocalhostSanitizer.sanitization(segmentChange);

client/src/main/java/io/split/client/LocalhostSplitChangeFetcher.java
```diff
@@ -23,9 +23,11 @@ public LocalhostSplitChangeFetcher(String filePath){
 
     @Override
     public SplitChange fetch(long since, FetchOptions options) {
+
         try {
-            JsonReader jsonReader = new JsonReader(new FileReader(_file));
-            return Json.fromJson(jsonReader, SplitChange.class);
+            ObjectMapper objectMapper = new ObjectMapper();
```
mmelograno, line 28: JsonReader jsonReader = new JsonReader(new FileReader(_file));
            SplitChange splitChange = Json.fromJson(jsonReader, SplitChange.class);
            return LocalhostSanitizer.sanitization(splitChange);

mmelograno (changes requested):

client/src/main/java/io/split/client/utils/LocalhostSanitizer.java
```diff
@@ -0,0 +1,144 @@
...
+                    split.conditions = new ArrayList<>();
+                }
+
+                Condition condition = new Condition();
+                if (!split.conditions.isEmpty()){
+                     condition = split.conditions.get(split.conditions.size() - 1);
+                }
+
+                if (split.conditions.isEmpty() || !condition.conditionType.equals(ConditionType.ROLLOUT) ||
+                        !condition.matcherGroup.matchers.get(0).matcherType.equals(MatcherType.ALL_KEYS)) {
```
mmelograno, line 80: missing matchers empty before accessing

mmelograno (changes requested):

client/pom.xml
nmayorsplit: We need that one to map the empty elements or elements that aren't.

client/src/main/java/io/split/client/LocalhostSegmentChangeFetcher.java
nmayorsplit: In this way, it's not going to work. If an element isn't in the JSON file is going to throw an exception.

client/src/main/java/io/split/client/utils/LocalhostSanitizer.java
nmayorsplit: I added the new check.

mmelograno (approved):

PR #385, opened 2023-04-03
Clean up classes that are no longer used. Fix LegacyLocalhostSplitChangeFetcher

nmayorsplit:


client/src/main/java/io/split/client/LegacyLocalhostSplitChangeFetcher.java
```diff
@@ -26,7 +26,7 @@ public class LegacyLocalhostSplitChangeFetcher implements SplitChangeFetcher {
     private final File _splitFile;
 
     public LegacyLocalhostSplitChangeFetcher(String directory) {
-        if (directory.equals("")){
+        if (directory == null || directory.equals("")){
```
gthea, line 29:
```suggestion
        if (directory == null || directory.isBlank()){
```

gthea (approved):

mmelograno (approved):

client/src/main/java/io/split/client/LegacyLocalhostSplitChangeFetcher.java
nmayorsplit: I updated but using isEmpty.

PR #393, opened 2023-05-05
[SDKS-6790] Update Default treatment to be control for yaml and json localhost

nmayorsplit:


client/src/main/java/io/split/client/YamlLocalhostSplitChangeFetcher.java
```diff
@@ -63,7 +63,7 @@ public SplitChange fetch(long since, FetchOptions options) {
                     split.conditions.add(condition);
                 }
                 split.status = Status.ACTIVE;
-                split.defaultTreatment = treatment;
+                split.defaultTreatment = "control";
                 split.trafficTypeName = "user";
                 split.trafficAllocation = 100;
                 split.trafficAllocationSeed = 1;
```
gthea, line 69: Maybe we can turn these values into constants in the future.

gthea (approved):

client/src/main/java/io/split/client/YamlLocalhostSplitChangeFetcher.java
nmayorsplit: I added LocalhostConstants that are used in some localhost classes.

PR #428, opened 2023-08-03
Shaded and non shaded versions

danieldbower:
Corrects the url of the sdk documentation as the given url does not work.

Minor fix added to specify the version number for the `maven-source-plugin` 

I also bumped the version to ensure I had some maven settings correct.

I was not able to run the redis module tests as that seems to be an integration test requiring some infrastructure.

The pom will now create a shaded version `java-client-with-deps` and a non-shaded version at `java-client`. 

In this version, documentation referencing the shaded version's maven coordinates would have to change from this:
``` xml
            <dependency>
                <version>4.9.0</version>
            </dependency>
```

to this

``` xml
            <dependency>
                <artifactId>java-client-with-deps</artifactId>
                <version>4.9.0</version>
            </dependency>
``` 

The original maven coordinates would be for the non-shaded version.

Why do this?
Simpler management of dependencies with tools like scanning for vulnerabilities and updates.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,6 @@
+4.9.0 (Aug 3, 2023)
+- Shuffled maven poms around to release both a client only jar and a client with dependencies version
+
 4.8.1 (Aug 1, 2023)
 - Applied linting rules to the code.
 - Fixed an issue when the prefix is empty for Redis settings.
--- a/client-with-deps/pom.xml
+++ b/client-with-deps/pom.xml
@@ -1,5 +1,6 @@
-<?xml version="1.0"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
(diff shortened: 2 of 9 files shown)


danieldbower:
While this works, I'm increasingly convinced that the correct way to do it is a separate module so that the dependency list in the pom is correct in both instances.

danieldbower:
Reasoning for the separate module:  https://www.mail-archive.com/users@maven.apache.org/msg139854.html

jbreton:
I'm using this package in a kotlin project. While it's working fine using JVM, it fails when using the native build (Graalvm). We believe the native build is not able to load some of the shaded dependencies. 
It's quite hard to prove since SplitClientImp says the sdk is ready but all calls to getTreatment [are failling](https://github.com/splitio/java-client/blob/9260637b2699081fb8942c89af2be61fcea19045/client/src/main/java/io/split/client/SplitClientImpl.java#L417). There are no other warnings or errors. It seems to at least call the API since we're getting a 400 Bad request error when using an invalid API Key.

danieldbower:
Are there any suggestions or concerns with this PR that would increase its chances of being merged?

danieldbower:
If you REALLY prefer to shade, then you should be prefixing those packages so they don't conflict with the rest of the app.  See this feature of the maven-shade-plugin:  https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html

agustinona:
Hi @danieldbower ,
We apologize for this not having gained traction yet, and thank you for the valuable contribution. Our team is currently analyzing this as it would mean supporting two packages along with the associated maintenance overhead.
We expect to have a resolution for this in the following quarters.

danieldbower:
I'd recommend that if you wanted to stick with just one artifact, you go with the non-shaded version.  That will remove some complexity and make the library more secure.
If you want to stay with the shaded version, please prefix those packages so they don't conflict with the rest of the app. See this feature of the maven-shade-plugin: https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html

silviuburceadev:
I feel like this should get some attention. More and more security issues appear for shaded libraries and instead of just bumping the versions in your own project, you have to wait for the fix in split.io Java client itself first, then bump its version to get the fix. I have personally opened 2 PRs in the last few weeks in this regard.
- https://github.com/splitio/java-client/pull/556
- https://github.com/splitio/java-client/pull/572
A solution to this, if you want to keep it like that, with shaded libs, would be to add renovate/dependabot to stay up to date with the fixes.

nmayorsplit:
Hi @danieldbower,
I've just released java-client with a non-shaded option. You can download the new version (4.18.3) and add a classifier option in your pom file, as shown in the example below:

```
<dependency>
            <version>4.18.3</version>
            <classifier>non-shaded</classifier>
  </dependency>
```
Please let me know if it works for you.
Thank you!

PR #429, opened 2023-08-30
[SDKS-7445] Add InputStream provider for localhost mode

nmayorsplit:


client/src/main/java/io/split/client/utils/InputStreamProviderImp.java
```diff
@@ -0,0 +1,18 @@
+package io.split.client.utils;
+
+import io.split.client.exceptions.InputStreamProviderException;
+
+import java.io.InputStream;
+
+public class InputStreamProviderImp implements InputStreamProvider {
+    private InputStream _inputStream;
```
gthea, line 8:
```suggestion
    private final InputStream _inputStream;
```

gthea (approved):

client/src/main/java/io/split/client/utils/InputStreamProviderImp.java
nmayorsplit: Updated.

mredolatti (approved):

PR #432, opened 2023-09-06
[SDKS-7514] Add flagSetFilter config

nmayorsplit:


client/src/main/java/io/split/client/SplitClientConfig.java
```diff
@@ -417,6 +424,7 @@ public static final class Builder {
         private StorageMode _storageMode = StorageMode.MEMORY;
         private final long _lastSeenCacheSize = 500000;
         private ThreadFactory _threadFactory;
+        private List<String> _flagSetsFilter = null;
```
mmelograno, line 427: empty will be better instead of null

mmelograno (changes requested):

client/src/main/java/io/split/client/SplitClientConfig.java
nmayorsplit: Updated.

sanzmauro (approved):

mmelograno (approved):

PR #435, opened 2023-09-08
[SDKS-7516] Create FlagSetSanitizer to validate flag sets

nmayorsplit:


client/src/main/java/io/split/client/utils/FlagSetSanitizer.java
```diff
@@ -0,0 +1,46 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public final class FlagSetSanitizer {
```
sanzmauro, line 12: please follow the spec and use the same name

client/src/main/java/io/split/client/utils/FlagSetSanitizer.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private static final String FLAG_SET_REGEX = "^[a-z0-9][_a-z0-9]{0,49}$";
+    private static final Logger _log = LoggerFactory.getLogger(FlagSetSanitizer.class);
+
+    private FlagSetSanitizer() {
+        throw new IllegalStateException("Utility class");
+    }
+
+    public static List<String> sanitizeFlagSet(List<String> flagSets) {
```
sanzmauro, line 21: same here

sanzmauro (changes requested):

client/src/main/java/io/split/client/utils/FlagSetSanitizer.java
nmayorsplit: Updated

client/src/main/java/io/split/client/utils/FlagSetSanitizer.java
nmayorsplit: Updated

sanzmauro (approved):

PR #437, opened 2023-09-12
[SDKS-7520] Add sets in HttpSplitChangeFetcher and FetchOptions

nmayorsplit:


client/src/main/java/io/split/client/HttpSplitChangeFetcher.java
```diff
@@ -97,6 +101,9 @@ public SplitChange fetch(long since, FetchOptions options) {
             }
 
             if (statusCode < HttpStatus.SC_OK || statusCode >= HttpStatus.SC_MULTIPLE_CHOICES) {
+                if (statusCode == HttpStatus.SC_REQUEST_URI_TOO_LONG) {
+                    _log.error("The amount of flag sets provided are big causing uri length error.");
+                }
                 _telemetryRuntimeProducer.recordSyncError(ResourceEnum.SPLIT_SYNC, statusCode);
                 _log.warn(String.format("Response status was: %s. Reason: %s", statusCode , response.getReasonPhrase()));
                 throw new IllegalStateException(String.format("Could not retrieve splitChanges since %s; http return code %s", since, statusCode));
```
chillaq, line 109: Are you passing the response.statusCode in the exception raised? I do see it only passed in the text message.
You will need the status code to stop retry synching

chillaq (changes requested):

client/src/main/java/io/split/client/HttpSplitChangeFetcher.java
nmayorsplit: Thank you Bilal, you are correct. So, I had to add a new exception to handle this case and create a new property in FetchResult to know if we should retry or not. All these changes are in this pr.

PR #439, opened 2023-09-19
[SDKS-7518] Add flagSets in InMemoryCacheImp

nmayorsplit:
- Implement GetNamesByFlagSets
- Update methods: update, remove, and add.

### DIFF
--- a/client/src/main/java/io/split/storages/memory/InMemoryCacheImp.java
+++ b/client/src/main/java/io/split/storages/memory/InMemoryCacheImp.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ConcurrentHashMultiset;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multiset;
 import com.google.common.collect.Sets;
+import io.split.client.interceptors.FlagSetsFilter;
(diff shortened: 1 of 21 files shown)


client/src/main/java/io/split/storages/memory/InMemoryCacheImp.java
```diff
@@ -160,4 +183,29 @@ public Set<String> getSegments() {
...
+                if (features == null) {
+                    features = new HashSet<>();
+                }
+                features.add(featureFlag.feature());
+                _flagSets.put(set, features);
+            }
+        }
+    }
+
+    private void removeFromFlagSets(String featureFlagName, HashSet<String> sets) {
```
EmilianoSanchez, line 204: can `sets` be null? check that case if necessary.

EmilianoSanchez (approved):

client/src/main/java/io/split/storages/memory/InMemoryCacheImp.java
nmayorsplit: It's true. So now we validate if not null, and I updated the test cases.

PR #440, opened 2023-09-22
[SDKS-7519] Add getNamesByFlagSets in UserCustomSplitAdapterConsumer

nmayorsplit:
- Add getMembers method in Pipeline and CustomStorage interfaces.

### DIFF
--- a/client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
+++ b/client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
@@ -94,8 +94,14 @@ public class UserCustomSplitAdapterConsumer  implements SplitCacheConsumer {
 
     @Override
     public Map<String, HashSet<String>> getNamesByFlagSets(List<String> flagSets) {
-        //Todo implement
-        return null;
+        Map<String, HashSet<String>> toReturn = new HashMap<>();
+        for (String set: flagSets) {
+            HashSet<String> keys = _userStorageWrapper.getMembers(PrefixAdapter.buildFlagSetPrefix(set));
+            if(keys != null){
+                toReturn.put(set, keys);
+            }
+        }
+        return toReturn;
     }
 
     @Override
--- a/client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
+++ b/client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
@@ -5,12 +5,14 @@ import io.split.client.utils.Json;
 import io.split.engine.experiments.ParsedSplit;
 import io.split.engine.experiments.SplitParser;
 import io.split.storages.SplitCacheConsumer;
+import io.split.storages.pluggable.domain.UserPipelineWrapper;
 import io.split.storages.pluggable.domain.UserStorageWrapper;
 import io.split.storages.pluggable.domain.PrefixAdapter;
 import io.split.storages.pluggable.utils.Helper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import pluggable.CustomStorageWrapper;
+import pluggable.Result;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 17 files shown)


client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
```diff
@@ -94,8 +96,20 @@ public List<String> splitNames() {
 
     @Override
     public Map<String, HashSet<String>> getNamesByFlagSets(List<String> flagSets) {
-        //Todo implement
-        return null;
+        Map<String, HashSet<String>> toReturn = new HashMap<>();
+        try {
+            UserPipelineWrapper pipelineExecution = _userStorageWrapper.pipeline();
+            for (String set: flagSets) {
```
gthea, line 102: Edge case of course but you could just return `toReturn` if `flagSets` is `null`, to avoid the exception.
```suggestion
        try {
            if (flagSets == null) {
                return toReturn;
            }
            UserPipelineWrapper pipelineExecution = _userStorageWrapper.pipeline();
            for (String set: flagSets) {
```

client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
```diff
@@ -94,8 +96,20 @@ public List<String> splitNames() {
...
+                 pipelineExecution.getMembers(PrefixAdapter.buildFlagSetPrefix(set));
+            }
+            List<Result> results = pipelineExecution.exec();
```
gthea, line 105: Not sure if `results` can ever be `null`; in that case maybe we could initialize `results` as empty list.

redis-wrapper/src/main/java/redis/RedisSingle.java
```diff
@@ -208,6 +208,17 @@ public List<String> getItems(List<String> keys) throws Exception {
         }
     }
 
+    @Override
+    public HashSet<String> getMembers(String key) throws Exception {
+        Set<String> flags;
```
gthea, line 213: Same as the previous comment
```suggestion
        Set<String> items;
```

redis-wrapper/src/main/java/redis/RedisCluster.java
```diff
@@ -231,6 +232,17 @@ public List<String> getItems(List<String> keys) throws Exception {
         }
     }
 
+    @Override
+    public HashSet<String> getMembers(String key) throws Exception {
+        Set<String> flags;
```
gthea, line 237: Maybe a more generic name would be more suitable?
```suggestion
        Set<String> items;
```

redis-wrapper/src/main/java/redis/RedisPipeline.java
```diff
@@ -29,6 +29,11 @@ public void hIncrement(String key, String field, long value) {
         _pipelined.hincrBy(_commonRedis.buildKeyWithPrefix(key), field, value);
     }
 
+    @Override
+    public void getMembers(String key) {
```
gthea, line 33: Isn't this one missing the `throws Exception` specified in the interface?

Also, I think `RedisException` instead of `Exception` might be enough

pluggable-storage/src/main/java/pluggable/CustomStorageWrapper.java
```diff
@@ -28,7 +29,8 @@ public interface CustomStorageWrapper {
     void addItems(String key, List<String> items) throws Exception;
     void removeItems(String key, List<String> items) throws Exception;
     List<String> getItems(List<String> keys) throws Exception;
+    HashSet<String> getMembers(String key) throws Exception;
```
gthea, line 32: I'd suggest using the `Set` interface instead, if possible
```suggestion
    Set<String> getMembers(String key) throws Exception;
```

client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
nmayorsplit: Updated.

client/src/main/java/io/split/storages/pluggable/adapters/UserCustomSplitAdapterConsumer.java
nmayorsplit: Updated. I added a check if it's null or empty.

redis-wrapper/src/main/java/redis/RedisSingle.java
nmayorsplit: Updated.

redis-wrapper/src/main/java/redis/RedisPipeline.java
nmayorsplit: This method is not going to throw an exception.

pluggable-storage/src/main/java/pluggable/CustomStorageWrapper.java
nmayorsplit: Updated.

PR #441, opened 2023-09-22
[SDKS-7523] Add sets in SplitView

nmayorsplit:


client/src/main/java/io/split/client/api/SplitView.java
```diff
@@ -24,13 +24,17 @@ public class SplitView {
...
     public Map<String, String> configs;
+    public List<String> sets = new ArrayList<>();
 
     public static SplitView fromParsedSplit(ParsedSplit parsedSplit) {
         SplitView splitView = new SplitView();
         splitView.name = parsedSplit.feature();
         splitView.trafficType = parsedSplit.trafficTypeName();
         splitView.killed = parsedSplit.killed();
         splitView.changeNumber = parsedSplit.changeNumber();
+        if (parsedSplit.flagSets() != null) {
```
mmelograno, line 35: correct me if I'm wrong but we will return empty not null

mmelograno (changes requested):

client/src/main/java/io/split/client/api/SplitView.java
nmayorsplit: Updated.

mmelograno (approved):

PR #444, opened 2023-09-29
[SDKS-7537] Add new latencies in MethodLatencies

nmayorsplit:


mmelograno (changes requested):
I would encourage you to follow JAVA naming patterns for the entire file. And I think you have missed pluggable and redis

mmelograno (approved):

PR #445, opened 2023-10-02
[SDKS-7539] Add new exceptions in MethodExceptions

nmayorsplit:


mmelograno (changes requested):
Same changes as previous PR as well as redis and pluggable coverage.

mmelograno (approved):

PR #446, opened 2023-10-04
[SDKS-7538] Add flag sets in Init Config

nmayorsplit:


client/src/main/java/io/split/inputValidation/FlagSetsValidator.java
```diff
@@ -31,19 +34,20 @@ public static HashSet<String> cleanup(List<String> flagSets) {
...
+                invalidSets ++;
                 _log.warn(String.format("you passed %s, Flag Set must adhere to the regular expressions %s. This means an Flag Set must be " +
                         "start with a letter, be in lowercase, alphanumeric and have a max length of 50 characters. %s was discarded.",
                         flagSet, FLAG_SET_REGEX, flagSet));
                 continue;
             }
             cleanFlagSets.add(flagSet);
         }
-        return cleanFlagSets;
+        return new FSValidatorResult(new ArrayList<>(cleanFlagSets), invalidSets);
```
EmilianoSanchez, line 45: I see that `cleanFlagSets` (a Set) is converted into a List here. And then, in line 51 it is converted into a Set again.
So maybe the first argument of `FSValidatorResult` could be a `Set<String>` to avoid the conversion Set->List->Set

EmilianoSanchez (approved):
LGTM, but there is a comment to check

client/src/main/java/io/split/inputValidation/FlagSetsValidator.java
nmayorsplit: I updated. Thank you Emi!

PR #450, opened 2023-10-20
[SDKS-7608] Fix null pointer when a feature flag has a set that is not in the storage

nmayorsplit:


client/src/main/java/io/split/storages/memory/InMemoryCacheImp.java
```diff
@@ -206,8 +206,10 @@ private void removeFromFlagSets(String featureFlagName, HashSet<String> sets) {
         if (sets != null) {
             for (String set : sets) {
                 HashSet<String> features = _flagSets.get(set);
-                features.remove(featureFlagName);
-                _flagSets.put(set, features);
+                if (features != null) {
+                    features.remove(featureFlagName);
+                    _flagSets.put(set, features);
```
EmilianoSanchez, line 211: Check Nadia, but maybe you can remove this line (`_flagSets.put(set, features);`), because `features` is a reference already inside the `_flagSets` ConcurrentMap.

EmilianoSanchez (approved):

client/src/main/java/io/split/storages/memory/InMemoryCacheImp.java
nmayorsplit: yes, you are right. I removed the extra line.

PR #455, opened 2023-10-25
FalgSets improvements and fixes

nmayorsplit:


client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -332,13 +334,16 @@ private Map<String, SplitResult> getTreatmentsWithConfigInternal(String matching
...
                 return new HashMap<>();
             }
-            featureFlagNames = getAllFlags(cleanFlagSets);
+            featureFlagNames = new ArrayList<>();
         } else if (featureFlagNames == null) {
             _log.error(String.format("%s: featureFlagNames must be a non-empty array", methodEnum.getMethod()));
             return new HashMap<>();
         }
         try {
             checkSDKReady(methodEnum, featureFlagNames);
```
sanzmauro, line 343: why are you not checking if the SDK is ready in the first line of this method? you are doing a lot of operations before checking the status, I think you can improve the "performance" doing the validations in the beginning.

ps: the method is really hard to read/understand because it has many `if else` conditions and to many lines.

nmayorsplit: I moved set validation to another method.

client/src/main/java/io/split/engine/common/SyncManagerImp.java
```diff
@@ -108,7 +109,7 @@ public static SyncManagerImp build(SplitTasks splitTasks,
                                                         config.getThreadFactory(),
                                                         splitParser,
                                                         splitCacheProducer,
-                                                        config.getSetsFilter());
+                                                        new FlagSetsFilterImpl(config.getSetsFilter()));
```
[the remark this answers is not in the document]
nmayorsplit: Updated.

sanzmauro (approved):

mmelograno (approved):

PR #464, opened 2023-12-01
Update version and changelogs

nmayorsplit:


CHANGES.txt
```diff
@@ -1,3 +1,6 @@
+4.10.2 (Dec 1, 2023)
+- Fixed some issues for flag sets.
```
EmilianoSanchez, line 2: Maybe you can be more specific with the issue that was fixed.
For example, `Fixed an issue for flag sets, that was causing the SDK to not log a warning when using flag sets that don't contain cached feature flags.`

EmilianoSanchez (approved):

emmaz90 (approved):

Issue #473, opened 2024-02-21
Deadlock issue when tracking impressions


ryaneorth:
We have a high throughput Spring-based web application that frequently encounters a deadlock scenario in the Split SDK, ultimately requiring restarting the web application to resolve.

Specifically, all threads get stuck with the following stack trace:

```
deployment.redrock.ear//io.split.client.impressions.UniqueKeysTrackerImp.track(UniqueKeysTrackerImp.java:59)
deployment.redrock.ear//io.split.client.impressions.strategy.ProcessImpressionNone.process(ProcessImpressionNone.java:28)
deployment.redrock.ear//io.split.client.impressions.ImpressionsManagerImpl.track(ImpressionsManagerImpl.java:116)
deployment.redrock.ear//io.split.client.SplitClientImpl.recordStats(SplitClientImpl.java:342)
deployment.redrock.ear//io.split.client.SplitClientImpl.getTreatmentWithConfigInternal(SplitClientImpl.java:258)
deployment.redrock.ear//io.split.client.SplitClientImpl.getTreatmentWithConfig(SplitClientImpl.java:99)
```

We encountered this a few months before and decided to switch to use `None` as our `impressionsMode` assuming that would solve the issue (we don't care about impressions anyway). However, digging into the code it appears the [ProcessImpressionsNone.java](https://github.com/splitio/java-client/blob/master/client/src/main/java/io/split/client/impressions/strategy/ProcessImpressionNone.java#L28) still uses the [UniqueKeysTracker](https://github.com/splitio/java-client/blob/master/client/src/main/java/io/split/client/impressions/UniqueKeysTrackerImp.java#L59) which is backed by a `ConcurrentHashMap`, ultimately deadlocking on the call to `put`.

To solve this issue, could we this issue be resolved in one of the two following ways:

1. Provide us with a true "no-op" impressions tracker that disables handling impressions entirely (e.g. the `process` method is empty)
2. Fix the deadlock issue within the UniqueKeysTracker - either by not using a `ConcurrentHashMap` or wrap the `put` call on the ConcurrentHashMap with some kind of timeout?

As mentioned, impressions are not incredibly important to us and the application not deadlocking is our biggest goal.

For what its worth, we are using version `4.7.0` of the client and looking at the changes since then I do not see any [changes](https://github.com/splitio/java-client/blob/master/CHANGES.txt#L1-L45) that would lead me to believe this issue has been addressed in the latest version (`4.11.0`).

[...]

PR #474, opened 2024-02-21
feat: add ProcessImpressionNoOp to ignore impressions entirely

ryaneorth:
Allows for bypassing impression tracking, solving the problem reported in #473


mmelograno:
Hi @ryaneorth , 

We cannot fully ignore usage metrics and unique keys since they are currently an important part of our business logic. Setting the impressions mode to None, is as low as we can go with outgoing traffic. Having said that, I’ll make sure this is tracked as a feature request. Meanwhile we released 4.11.1 which includes a revamp of None mode logic.

Thanks,
Matias

ryaneorth:
Great, thank you Matias!

PR #478, opened 2024-03-12
added httpclient wrapper

chillaq:


mmelograno (changes requested):
An interface contract will be useful that forces you with the methods Get Post for now more generic.

client/src/main/java/io/split/client/SplitHttpClient.java
```diff
@@ -0,0 +1,18 @@
...
+import io.split.engine.common.FetchOptions;
+import io.split.telemetry.domain.enums.HttpParamsWrapper;
+
+import org.apache.hc.core5.http.HttpEntity;
+import java.io.IOException;
+import java.net.URI;
+import java.util.Map;
+
+public interface SplitHttpClient {
+    public String get(URI uri, FetchOptions options, HttpParamsWrapper telemetryParamsWrapper);
```
mmelograno, line 12: preferred tu return some type of dto on both calls with statusCode, statusMessage, body so HTTPClient only performs Get and Post then each recorder will decide to track telemetry or whatever it needs to handle

mmelograno (changes requested):

client/src/main/java/io/split/client/SplitHttpClient.java
chillaq: will do

mmelograno (approved):

PR #480, opened 2024-03-14
Custom auth segmentfetcher

chillaq:


client/src/main/java/io/split/client/HttpSegmentChangeFetcher.java
```diff
@@ -69,42 +66,27 @@ public SegmentChange fetch(String segmentName, long since, FetchOptions options)
...
-            if(options.cacheControlHeadersEnabled()) {
-                request.setHeader(CACHE_CONTROL_HEADER_NAME, CACHE_CONTROL_HEADER_VALUE);
-            }
-
-            response = _client.execute(request);
-
-            int statusCode = response.getCode();
-
-            if (_log.isDebugEnabled()) {
-                _log.debug(String.format("[%s] %s. Status code: %s", request.getMethod(), uri.toURL(), statusCode));
```
nmayorsplit, line 83: For me, we don't should remove this log. You can leave this log on the debug level to know what happens when we fetch segments.

nmayorsplit (changes requested):

client/src/main/java/io/split/client/HttpSegmentChangeFetcher.java
chillaq: The log is in the SplitHttpClientImpl class

client/src/main/java/io/split/client/HttpSegmentChangeFetcher.java
```diff
@@ -37,16 +34,16 @@ public final class HttpSegmentChangeFetcher implements SegmentChangeFetcher {
     private static final String CACHE_CONTROL_HEADER_NAME = "Cache-Control";
     private static final String CACHE_CONTROL_HEADER_VALUE = "no-cache";
```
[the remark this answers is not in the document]
chillaq: will do

nmayorsplit (approved):
LGTM

PR #482, opened 2024-03-18
Updated EventSender and HttpPostImp classes

chillaq:


client/src/main/java/io/split/client/events/EventsSender.java
```diff
@@ -4,6 +4,7 @@
 import io.split.client.dtos.Event;
 import io.split.client.utils.Utils;
 import io.split.service.HttpPostImp;
+import io.split.service.SplitHttpClient;
 import io.split.telemetry.domain.enums.HttpParamsWrapper;
 import io.split.telemetry.storage.TelemetryRuntimeProducer;
 import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
```
nmayorsplit, line 10: Please, remove this import that we aren't using anymore.

client/src/main/java/io/split/service/HttpPostImp.java
```diff
@@ -1,5 +1,6 @@
 package io.split.service;
 
+import io.split.client.dtos.SplitHttpResponse;
 import io.split.client.utils.Utils;
 import io.split.telemetry.domain.enums.HttpParamsWrapper;
 import io.split.telemetry.storage.TelemetryRuntimeProducer;
```
nmayorsplit, line 6: Also here, remove CloseableHttpClient and CloseableHttpResponse imports.

client/src/main/java/io/split/service/SplitHttpClientImpl.java
```diff
@@ -33,7 +33,7 @@ public static SplitHttpClientImpl create(
         return new SplitHttpClientImpl(client, requestDecorator);
     }
 
-    private SplitHttpClientImpl
+    public SplitHttpClientImpl
```
nmayorsplit, line 36: Why did you make this method public? You can use the create method.

nmayorsplit (changes requested):

client/src/main/java/io/split/client/events/EventsSender.java
chillaq: will do

client/src/main/java/io/split/service/SplitHttpClientImpl.java
chillaq: Some tests create the httpclient outside methods, which require try catch block, so I made it public, I will updated these tests and revert back to private.

PR #490, opened 2024-03-26
Release 4.12.0-rc1

chillaq:


client/src/main/java/io/split/client/dtos/SplitHttpResponse.java
```diff
@@ -0,0 +1,10 @@
+package io.split.client.dtos;
+
+/**
+ * A structure for returning http call results information
+ */
+public class SplitHttpResponse {
+    public Integer statusCode;
+    public String statusMessage;
+    public String body;
```
mredolatti, line 9: can we add response headers as well? Also, how about making this class immutable? so that no manipulation can take place and it becomes inherently thread-safe

client/src/main/java/io/split/client/impressions/HttpImpressionsSender.java
```diff
@@ -63,32 +63,24 @@ private HttpImpressionsSender(CloseableHttpClient client, URI impressionBulkTarg
...
     public void postImpressionsBulk(List<TestImpressions> impressions) {
 
-        CloseableHttpResponse response = null;
+        SplitHttpResponse response;
         long initTime = System.currentTimeMillis();
         try {
             HttpEntity entity = Utils.toJsonEntity(impressions);
+            Map<String, String> additionalHeader = new HashMap<>();
+            additionalHeader.put(IMPRESSIONS_MODE_HEADER, _mode.toString());
+            response = _client.post(_impressionBulkTarget, entity, additionalHeader);
```
mredolatti, line 72: declare response here

client/src/main/java/io/split/service/SplitHttpClient.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Map;
+
+public interface SplitHttpClient {
+    /**
+     * Wrapper for HTTP get method
+     * @param uri the URL to be used
+     * @param options The FetchOptions object that contains headers.
+     * @return The response structure SplitHttpResponse
+     */
+    public SplitHttpResponse get(URI uri, FetchOptions options);
```
mredolatti, line 19: should we allow custom headers here as well?

mredolatti (changes requested):

client/src/main/java/io/split/client/HttpSplitChangeFetcher.java
```diff
@@ -75,38 +68,24 @@ public SplitChange fetch(long since, FetchOptions options) {
                 uriBuilder.addParameter(SETS, "" + options.flagSetsFilter());
             }
             URI uri = uriBuilder.build();
+            response = _client.get(uri, options);
```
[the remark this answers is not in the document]
chillaq: will do

client/src/main/java/io/split/client/RequestDecorator.java
```diff
@@ -0,0 +1,60 @@
...
+                ? new NoOpHeaderDecorator()
+                : headerDecorator;
+    }
+
+    public HttpUriRequestBase decorateHeaders(HttpUriRequestBase request) {
+        try {
+            Map<String, String> headers = _headerDecorator.getHeaderOverrides();
+            for (Map.Entry entry : headers.entrySet()) {
+                if (isHeaderAllowed(entry.getKey().toString())) {
+                    request.addHeader(entry.getKey().toString(), entry.getValue());
```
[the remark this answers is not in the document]
chillaq: when using addHeader, there will be only one  Authorization header with multiple values separated by comma.  This does not impact the proxy auth since we support "Basic xxxxxx" which is different than the default "Bearer xxxxx"

client/src/main/java/io/split/client/impressions/HttpImpressionsSender.java
chillaq: will do

client/src/main/java/io/split/service/SplitHttpClient.java
chillaq: makes sense, will do

client/src/main/java/io/split/client/dtos/SplitHttpResponse.java
chillaq: will do

mredolatti (changes requested):

client/src/main/java/io/split/client/UserCustomHeaderDecorator.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.Map;
+
+public interface UserCustomHeaderDecorator
+{
+    /**
+     * Get the additional headers needed for all http operations
+     * @return HashMap of addition headers
+     */
+    Map<String, String> getHeaderOverrides();
```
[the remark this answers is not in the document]
chillaq: will do

PR #497, opened 2024-04-23
added semver class

chillaq:


client/src/main/java/io/split/engine/matchers/Semver.java
```diff
@@ -0,0 +1,161 @@
...
+    private void setMajorMinorAndPatch(String version) throws SemverParseException {
+        String[] vParts = version.split(ValueDelimiter);
+        if (vParts.length != 3)
+            throw new SemverParseException("Unable to convert to Semver, incorrect format: " + version);
+        _major = Long.parseUnsignedLong(vParts[0]);
+        _minor = Long.parseUnsignedLong(vParts[1]);
+        _patch = Long.parseUnsignedLong(vParts[2]);
+    }
+
+    private String setVersion() {
```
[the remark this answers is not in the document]
chillaq: will update and add a test for it.

client/src/main/java/io/split/engine/matchers/Semver.java
```diff
@@ -0,0 +1,178 @@
...
+        // Compare pre-release versions lexically
+        int minLength = Math.min(_preRelease.length, toCompare.PreRelease().length);
+        for (int i = 0; i < minLength; i++) {
+            if (_preRelease[i].equals(toCompare.PreRelease()[i])) {
+                continue;
+            }
+            if ( Ints.tryParse(_preRelease[i]) != null &&  Ints.tryParse(toCompare._preRelease[i]) != null) {
+                return Integer.compare(Integer.parseInt(_preRelease[i]), Integer.parseInt(toCompare._preRelease[i]));
+            }
+            return AdjustNumber(_preRelease[i].compareTo(toCompare._preRelease[i]));
```
sanzmauro, line 101: why do you need the AdjustNumber?

chillaq: The compare function returns numbers higher than 1, so I need to adjust it to return -1, 0 or 1

PR #499, opened 2024-04-26
Semver greater equalto matcher

chillaq:


client/src/main/java/io/split/engine/matchers/GreaterThanOrEqualToSemverMatcher.java
```diff
@@ -0,0 +1,54 @@
...
+        Semver matchSemver = Semver.build(matchValue.toString());
+        if (matchSemver == null) {
+            return false;
+        }
+
+        return _semVer != null && matchSemver.Compare(_semVer) >= 0;
+    }
+
+    @Override
+    public String toString() {
```
sanzmauro, line 29: why do you need this?

client/src/main/java/io/split/engine/matchers/GreaterThanOrEqualToSemverMatcher.java
```diff
@@ -0,0 +1,54 @@
...
+        StringBuilder bldr = new StringBuilder();
+        bldr.append("== ");
+        bldr.append(_semVer);
+        return bldr.toString();
+    }
+
+    @Override
+    public int hashCode() {
```
sanzmauro, line 37: and this?

client/src/main/java/io/split/engine/matchers/GreaterThanOrEqualToSemverMatcher.java
```diff
@@ -0,0 +1,54 @@
...
+        int result = 17;
+        result = 31 * result + _semVer.hashCode();
+        return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
sanzmauro, line 44: and this

sanzmauro (changes requested):

client/src/main/java/io/split/engine/matchers/GreaterThanOrEqualToSemverMatcher.java
chillaq: Other matcher classes has this implementation, I added it to avoid breaking anything else.

client/src/main/java/io/split/engine/matchers/GreaterThanOrEqualToSemverMatcher.java
chillaq: Other matcher classes has this implementation, I added it to avoid breaking anything else.

client/src/main/java/io/split/engine/matchers/GreaterThanOrEqualToSemverMatcher.java
chillaq: Other matcher classes has this implementation, I added it to avoid breaking anything else.

client/src/main/java/io/split/engine/matchers/GreaterThanOrEqualToSemverMatcher.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final Semver _semVer;
+
+    public GreaterThanOrEqualToSemverMatcher(String semVer) {
+        _semVer = Semver.build(semVer);
+    }
+
+    @Override
+    public boolean match(Object matchValue, String bucketingKey, Map<String, Object> attributes, EvaluationContext evaluationContext) {
+        if (matchValue == null) {
```
sanzmauro, line 17: same here, check if `_semVer` is null in this if.

chillaq: done

PR #500, opened 2024-04-26
added less than or equal semver matcher

chillaq:


client/src/main/java/io/split/engine/matchers/LessThanOrEqualToSemverMatcher.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final Semver _semVer;
+
+    public LessThanOrEqualToSemverMatcher(String semVer) {
+        _semVer = Semver.build(semVer);
+    }
+
+    @Override
+    public boolean match(Object matchValue, String bucketingKey, Map<String, Object> attributes, EvaluationContext evaluationContext) {
+        if (matchValue == null) {
```
sanzmauro, line 17: you can check here if `_semVer != null`

sanzmauro (changes requested):

client/src/main/java/io/split/engine/matchers/LessThanOrEqualToSemverMatcher.java
chillaq: will update

sanzmauro (approved):

PR #505, opened 2024-05-09
Fix empty token handling in RawAuthResponse constructor

hpark-miovision:
RawAuthResponse constructor first checks to ensure provided token is not null and is not empty. Existing implementation used != operator to detect non-empty string, thus failed to handle when provided token is actually empty, i.e. "".

The token comparison logic in RawAuthResponse constructor is now updated to use isEmpty() instead, more specifically, (token != null && !token.isEmpty()) to properly handle "" case.

New test case with sample json has been added for validation; the test prior to the change would fail with java.lang.ArrayIndexOutOfBoundsException.


nmayorsplit (approved):

nmayorsplit:
Thank you @hpark-miovision for the suggestion.
We have released Java 4.12.0 and this change was added.

Thanks,

Nadia

hpark-miovision:
Thank you very much @nmayorsplit for the update! We're using the splitio/java-client for integration with Split Proxy (https://help.split.io/hc/en-us/articles/4415960499213-Split-Proxy), and newer version should allow us to utilize auth endpoint of proxy as well.

PR #511, opened 2024-08-13
Kerberos support

chillaq:


client/src/main/java/io/split/client/SplitClientConfig.java
```diff
@@ -91,6 +92,7 @@ public class SplitClientConfig {
     private final HashSet<String> _flagSetsFilter;
     private final int _invalidSets;
     private final CustomHeaderDecorator _customHeaderDecorator;
+    private final String _authScheme;
```
sanzmauro, line 95: Please create an enum for this property which accepts `kerberos`

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
```diff
@@ -0,0 +1,220 @@
...
+import java.io.*;
+import java.net.*;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class SplitHttpClientKerberosImpl implements SplitHttpClient {
+
+    private static final Logger _log = LoggerFactory.getLogger(SplitHttpClient.class);
```
nmayorsplit, line 26:
```suggestion
    private static final Logger _log = LoggerFactory.getLogger(SplitHttpClientKerberosImpl.class);
```
Please update the class to get a logger.

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+    public SplitHttpResponse _get(HttpURLConnection getHttpURLConnection, FetchOptions options, Map<String, List<String>> additionalHeaders) {
+        InputStreamReader inputStreamReader = null;
+        try {
+            getHttpURLConnection.setRequestMethod("GET");
+
+            setBasicHeaders(getHttpURLConnection);
+            setAdditionalAndDecoratedHeaders(getHttpURLConnection, additionalHeaders);
+
+            if (options.cacheControlHeadersEnabled()) {
```
sanzmauro, line 78: what happen if `options.targetCN()` is != `null` ? are you ignoring him? 

or if `options.hasCustomCN()`  is true?

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
```diff
@@ -0,0 +1,220 @@
...
+            try {
+                if (getHttpURLConnection != null) {
+                    getHttpURLConnection.disconnect();
+                }
+            } catch (Exception e) {
+                _log.error(String.format("Could not close HTTP URL Connection: %s", e), e);
+            }
+        }
+    }
+    public SplitHttpResponse _get(HttpURLConnection getHttpURLConnection, FetchOptions options, Map<String, List<String>> additionalHeaders) {
```
nmayorsplit, line 70:
```suggestion
    public SplitHttpResponse get(HttpURLConnection getHttpURLConnection, FetchOptions options, Map<String, List<String>> additionalHeaders) {
```

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
```diff
@@ -0,0 +1,220 @@
...
+
+    private void setAdditionalAndDecoratedHeaders(HttpURLConnection urlConnection, Map<String, List<String>> additionalHeaders) {
+        if (additionalHeaders != null) {
+            for (Map.Entry<String, List<String>> entry : additionalHeaders.entrySet()) {
+                for (String value : entry.getValue()) {
+                    urlConnection.setRequestProperty(entry.getKey(), value);
+                }
+            }
+        }
+        HttpRequest request = new HttpGet("");
```
sanzmauro, line 196: why do you need to create a request?

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
```diff
@@ -0,0 +1,220 @@
...
+                BasicHeader responseHeader = new BasicHeader(entry.getKey(), entry.getValue());
+                responseHeaders.add(responseHeader);
+            }
+        }
+        return responseHeaders.toArray(new Header[0]);
+
+    }
+    @Override
+    public void close() throws IOException {
+
```
sanzmauro, line 218: Is it ok that it's empty?

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
```diff
@@ -0,0 +1,220 @@
...
+                if (postHttpURLConnection != null) {
+                    postHttpURLConnection.disconnect();
+                }
+            } catch (Exception e) {
+                _log.error(String.format("Could not close URL Connection: %s", e), e);
+            }
+        }
+    }
+
+    public SplitHttpResponse _post(HttpURLConnection postHttpURLConnection,
```
nmayorsplit, line 142:
```suggestion
    public SplitHttpResponse post(HttpURLConnection postHttpURLConnection,
```

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
```diff
@@ -0,0 +1,220 @@
...
+                    urlConnection.setRequestProperty(entry.getKey(), value);
+                }
+            }
+        }
+        HttpRequest request = new HttpGet("");
+        _requestDecorator.decorateHeaders(request);
+        for (Header header : request.getHeaders()) {
+            urlConnection.setRequestProperty(header.getName(), header.getValue());
+        }
+        request = null;
```
nmayorsplit, line 201:
```suggestion

```

nmayorsplit (commented):
There are some things to update.

client/src/main/java/io/split/client/SplitClientConfig.java
chillaq: There is only one value for now, is it worth enum?

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
chillaq: Yes, this is for compatibility reason

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
chillaq: will do

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
chillaq: This is an internal call that takes HttpURLConnection to make the testing easier.
There is another get() call which takes URI as the first parameter

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
chillaq: will update

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
chillaq: This is an internal call that takes HttpURLConnection to make the testing easier.
There is another get() call which takes URI as the first parameter

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
chillaq: This is the same logic in the SplitHttpClientImpl, this option is checked on the Segment and Split Fetcher classes

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
nmayorsplit: So, can we find a better name instead of an underscore?

client/src/main/java/io/split/client/SplitClientConfig.java
chillaq: ok, will do

client/src/main/java/io/split/service/SplitHttpClientKerberosImpl.java
chillaq: Sure, will do

client/src/main/java/io/split/client/SplitClientConfig.java
```diff
@@ -960,6 +968,17 @@ public Builder customHeaderDecorator(CustomHeaderDecorator customHeaderDecorator
             return this;
         }
 
+        /**
+         * Authentication Scheme
+         *
+         * @param authScheme
+         * @return this builder
+         */
+        public Builder authScheme(String authScheme) {
```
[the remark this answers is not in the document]
chillaq: will do

client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -525,6 +526,13 @@ private static SplitHttpClient buildSplitHttpClient(String apiToken, SplitClient
             httpClientbuilder = setupProxy(httpClientbuilder, config);
         }
 
+        if (config.authScheme() != null) {
```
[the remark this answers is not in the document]
chillaq: will do

sanzmauro (changes requested):
You have 7 code smells. Please try to fix them before merge 🙏 

Have you already reviewed them?

chillaq:
> You have 7 code smells. Please try to fix them before merge 🙏
> 
> Have you already reviewed them?

I reviewed and fixed all except one, cannot use try-with-reference since the variable InputStreamReader cannot be final.

sanzmauro (approved):

PR #523, opened 2024-12-12
Impression toggle

chillaq:


NicoZelaya (approved):

client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -634,9 +634,11 @@ private ImpressionsManagerImpl buildImpressionsManager(SplitClientConfig config,
         ImpressionListener listener = !impressionListeners.isEmpty()
                 ? new ImpressionListener.FederatedImpressionListener(impressionListeners)
                 : null;
+        counter = new ImpressionCounter();
```
nmayorsplit, line 637: Can we move this when it is NONE case?

nmayorsplit (changes requested):

client/src/main/java/io/split/client/SplitFactoryImpl.java
chillaq: We have to create the None strategy by default now, which is why we have the counter object as well.

nmayorsplit: Good, I got it. Thanks Bilal.

nmayorsplit (approved):

PR #537, opened 2025-03-27
Imp pr client

chillaq:


client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -123,111 +132,277 @@ public Map<String, String> getTreatments(String key, List<String> featureFlagNam
...
+    }
+
+    @Override
+    public String getTreatment(String key, String featureFlagName, EvaluationOptions evaluationOptions) {
+        return getTreatment(key, featureFlagName, Collections.<String, Object>emptyMap(), evaluationOptions);
+    }
+
+    @Override
+    public String getTreatment(String key, String featureFlagName, Map<String, Object> attributes, EvaluationOptions evaluationOptions) {
+        return getTreatmentWithConfigInternal(key, null, featureFlagName, attributes, evaluationOptions, MethodEnum.TREATMENT).treatment();
```
sanzmauro, line 251: why are you using `getTreatmentWithConfigInternal` here and `getTreatment` in line 246? what is the difference?

client/src/main/java/io/split/client/SplitClientImpl.java
```diff
@@ -373,8 +548,19 @@ private SplitResult getTreatmentWithConfigInternal(String matchingKey, String bu
...
+    private String validateProperties(Map<String, Object> properties) {
+        String validatedProperties = null;
+        if (properties != null) {
+            ImpressionPropertiesValidator.ImpressionPropertiesValidatorResult iPValidatorResult = ImpressionPropertiesValidator.propertiesAreValid(
+                    properties);
+            validatedProperties =  new GsonBuilder().create().toJson(iPValidatorResult.getValue()).toString();
+        }
+        return validatedProperties;
+    }
+
```
sanzmauro, line 560:
```suggestion
    private String validateProperties(Map<String, Object> properties) {
         if (properties == null){
             return null;
         }
        
         ImpressionPropertiesValidator.ImpressionPropertiesValidatorResult iPValidatorResult = ImpressionPropertiesValidator.propertiesAreValid(
                    properties);
         return  new GsonBuilder().create().toJson(iPValidatorResult.getValue()).toString();
    }

```

client/src/main/java/io/split/inputValidation/ImpressionPropertiesValidator.java
```diff
@@ -0,0 +1,83 @@
...
+                        + "Current one is %s bytes. Properties field is ignored", size));
+
+                return new ImpressionPropertiesValidatorResult(false);
+            }
+
+            result.put(entry.getKey(), value);
+        }
+
+        return new ImpressionPropertiesValidatorResult(true, size, result);
+    }
```
sanzmauro, line 52: These validations are exactly the same as the ones [here](https://github.com/splitio/java-client/blob/master/client/src/main/java/io/split/inputValidation/EventsValidator.java), if I'm correct. Can we improve this implementation and avoid duplicating code?

sanzmauro (changes requested):
did you validate these changes with live tail??

client/src/main/java/io/split/client/SplitClient.java
```diff
@@ -462,6 +463,488 @@ public interface SplitClient {
...
+     * <li>The key did not match any of the conditions in the feature flag roll-out plan</li>
+     * </ol>
+     * The default treatment of a feature flag is set on the Split user interface.
+     * <p/>
+     * <p/>
+     * This method does not throw any exceptions. It also never returns null.
+     *
+     * @param key a unique key of your customer (e.g. user_id, user_email, account_id, etc.) MUST not be null.
+     * @param featureFlagName the name of the feature flag we want to evaluate. MUST NOT be null.
+     * @param properties a json structure to attach to the impression.
```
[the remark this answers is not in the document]
chillaq: will correct

client/src/main/java/io/split/client/SplitClientImpl.java
chillaq: The getTreatment in 246 does not have attributes param, it is calling getTreatment in 250 with attributes null.

chillaq:
> did you validate these changes with live tail??

Not yet, validated only in integration tests

client/src/main/java/io/split/inputValidation/ImpressionPropertiesValidator.java
chillaq: will do

sanzmauro (approved):

PR #540, opened 2025-04-05
Added inmemory storage classes

chillaq:


client/src/main/java/io/split/engine/experiments/ParsedRuleBasedSegment.java
```diff
@@ -0,0 +1,125 @@
...
+                .collect(Collectors.toSet());
+    }
+
+    private static boolean isSegmentMatcher(AttributeMatcher attributeMatcher) {
+        return ((AttributeMatcher.NegatableMatcher) attributeMatcher.matcher()).delegate() instanceof UserDefinedSegmentMatcher;
+    }
+
+    private static UserDefinedSegmentMatcher asSegmentMatcherForEach(AttributeMatcher attributeMatcher) {
+        return (UserDefinedSegmentMatcher) ((AttributeMatcher.NegatableMatcher) attributeMatcher.matcher()).delegate();
+    }
```
sanzmauro, line 123: what is this for? 🤔

client/src/main/java/io/split/storages/RuleBasedSegmentCacheCommons.java
```diff
@@ -0,0 +1,8 @@
+package io.split.storages;
+
+import java.util.Set;
+
+public interface RuleBasedSegmentCacheCommons {
```
sanzmauro, line 5: why do you need this class?

client/src/main/java/io/split/storages/memory/RuleBasedSegmentCacheInMemoryImp.java
```diff
@@ -0,0 +1,101 @@
...
+
+    public void clear() {
+        _concurrentMap.clear();
+    }
+
+    public void putMany(List<ParsedRuleBasedSegment> ruleBasedSegments) {
+        for (ParsedRuleBasedSegment ruleBasedSegment : ruleBasedSegments) {
+            _concurrentMap.put(ruleBasedSegment.ruleBasedSegment(), ruleBasedSegment);
+        }
+    }
```
sanzmauro, line 82: why this method is public?

sanzmauro (changes requested):

client/src/main/java/io/split/engine/experiments/ParsedRuleBasedSegment.java
chillaq: These methods are  to check if the attribute is UserDefinedSegmentMatcher type, they exist in the other matchers that use the same attribute type

client/src/main/java/io/split/storages/memory/RuleBasedSegmentCacheInMemoryImp.java
chillaq: Its used for tests

sanzmauro (approved):

PR #541, opened 2025-04-07
Updated Evaluator

chillaq:


client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -214,6 +218,7 @@ public SplitFactoryImpl(String apiToken, SplitClientConfig config) throws URISyn
         // Segments
         _segmentSynchronizationTaskImp = buildSegments(config, segmentCache, splitCache);
 
+
```
sanzmauro, line 221:
```suggestion
```

sanzmauro (changes requested):
please add UT for RuleBasedSegmentMatcher class

PR #548, opened 2025-04-18
Added RBS support to localhost

chillaq:


client/src/main/java/io/split/client/JsonLocalhostSplitChangeFetcher.java
```diff
@@ -24,49 +24,61 @@ public class JsonLocalhostSplitChangeFetcher implements SplitChangeFetcher {
...
+        lastHashRuleBasedSegments = currHashRuleBasedSegments;
         splitChangeToProcess.featureFlags.s = changeNumber;
+        splitChangeToProcess.ruleBasedSegments.s = changeNumberRBS;
+
         return splitChangeToProcess;
     }
+
+    private <T> boolean checkExitConditions(ChangeDto<T> change, long cn) {
+        return change.t < cn && change.t != -1;
+    }
```
sanzmauro, line 75: we are repeating this validation in more than one place. I suggest to moving it to a utility

sanzmauro (approved):
i left a minor comment.

client/src/main/java/io/split/client/JsonLocalhostSplitChangeFetcher.java
chillaq: will fix

PR #586, opened 2025-07-16
Added new proxy config params

chillaq:


client/src/main/java/io/split/client/SplitFactoryImpl.java
```diff
@@ -626,22 +628,43 @@ private static SSLContext buildSSLContext(SplitClientConfig config) throws IOExc
...
         httpClientbuilder.setRoutePlanner(routePlanner);
 
-        if (config.proxyUsername() != null && config.proxyPassword() != null) {
+        if ((config.proxyUsername() != null && config.proxyPassword() != null) ||
+                (config.proxyConfiguration() != null && config.proxyConfiguration().getProxyCredentialsProvider() != null &&
+                        config.proxyConfiguration().getProxyCredentialsProvider() instanceof io.split.client.dtos.BasicCredentialsProvider)) {
             _log.debug("Proxy setup using credentials");
+            String userName;
+            String password;
+            if (config.proxyUsername() != null && config.proxyPassword() != null) {
```
EmilianoSanchez, line 646: Check that if both the old options and new Proxy configuration options are provided, the old options are ignored and a warning should be logged. Maybe something like:

`"[WARN] `proxyHost`, `proxyPort`, `proxyUsername` and `proxyPassword` configuration methods are deprecated. Please use the `ProxyConfiguration` object instead."`

EmilianoSanchez (changes requested):

client/src/main/java/io/split/client/SplitFactoryImpl.java
chillaq: will do

EmilianoSanchez (changes requested):

PR #614, opened 2025-12-19
feat: Add GraalVM native-image support by removing Guava shading

alexcol23:
## Description

This PR enables GraalVM native-image compilation for applications using the Split.io Java SDK.

## Problem

The Split SDK shades Guava to `split.com.google.common.*` to avoid dependency conflicts. However, this shaded Guava lacks the GraalVM native-image configuration that standard Guava provides, causing build failures:

```
Fatal error: com.oracle.graal.pointsto.util.AnalysisError$ParsingError: 
Error encountered while parsing split.com.google.common.cache.LocalCache$Segment.insertLoadingValueReference(LocalCache.java:2411)

Caused by: com.oracle.graal.pointsto.constraints.UnresolvedElementException: 
Discovered unresolved type during parsing: split.com.google.common.util.concurrent.SettableFuture
```

## Root Cause

Standard Guava (`com.google.common.*`) includes GraalVM native-image configuration in `META-INF/native-image/`. When the SDK shades Guava to `split.com.google.common.*`, this configuration is lost, and GraalVM cannot parse complex classes like `LocalCache.Segment`.

## Solution

Remove Guava from the shade plugin configuration. Guava remains as a transitive dependency, allowing applications to use the standard Guava library which has proper GraalVM native-image support.

## Changes

- Remove `com.google.guava:guava` from shade plugin includes
- Remove `org.checkerframework:*` from shade plugin includes  
- Remove `com.google` → `split.com.google` relocation
- Remove `org.checkerframework` → `split.org.checkerframework` relocation

## Testing

- Built native image with Quarkus 3.27.1 + Mandrel JDK-23
- Verified Split client works correctly at runtime
- Tested feature flag resolution in native image
- Native image startup time: ~0.5s (vs ~5s with JVM)

### Breaking Change
Applications that explicitly depend on the shaded Guava package names (`split.com.google.*`) will need to update imports. However, this is unlikely since the shaded packages are internal implementation details.

### Benefits
- Enables GraalVM native-image support for all Split SDK users
- Allows use with Quarkus, Micronaut, and other native-image frameworks
- No functional changes to the SDK behavior

### Potential Risks
- Version conflicts with other Guava users (mitigated by dependency management)
- The Split SDK uses stable Guava APIs (Cache, BloomFilter) that rarely change

## Related Frameworks

This fix enables Split SDK usage with:
- [Quarkus](https://quarkus.io/) - Supersonic Subatomic Java
- [Micronaut](https://micronaut.io/) - Modern JVM-based framework
[...]

PR #617, opened 2026-01-14
feat: Add GraalVM native-image support by removing Guava shading

alexcol23:
This commit enables GraalVM native-image compilation for applications using the Split.io Java SDK.

Problem:
The Split SDK shades Guava to split.com.google.common.* to avoid dependency conflicts. However, this shaded Guava lacks the GraalVM native-image configuration that standard Guava provides, causing build failures with errors like:
  'Discovered unresolved type: split.com.google.common.util.concurrent.SettableFuture'

Root Cause:
Standard Guava (com.google.common.*) includes GraalVM native-image configuration in META-INF/native-image/. When the SDK shades Guava to split.com.google.common.*, this configuration is lost.

Solution:
Remove Guava from the shade plugin configuration. Guava remains as a transitive dependency, allowing applications to use the standard Guava library which has proper GraalVM native-image support.

Changes:
- Remove com.google.guava:guava from shade plugin includes
- Remove org.checkerframework:* from shade plugin includes
- Remove com.google -> split.com.google relocation
- Remove org.checkerframework -> split.org.checkerframework relocation
- Update version to 4.18.2-graalvm to distinguish from upstream

Testing:
- Built native image with Quarkus 3.27.1 + Mandrel JDK-23
- Verified Split client works correctly at runtime
- Tested feature flag resolution in native image

Reference: https://github.com/splitio/java-client/pull/614

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -5,9 +5,9 @@
     <parent>
         <groupId>io.split.client</groupId>
         <artifactId>java-client-parent</artifactId>
-        <version>4.18.2</version>
+        <version>4.18.2-graalvm</version>
     </parent>
-    <version>4.18.2</version>
+    <version>4.18.2-graalvm</version>
     <artifactId>java-client</artifactId>
     <packaging>jar</packaging>
     <name>Java Client</name>
@@ -70,11 +70,13 @@
                                     <include>org.apache.httpcomponents.*</include>
                                     <include>org.apache.hc.*</include>
                                     <include>com.google.code.gson:gson</include>
-                                    <include>com.google.guava:guava</include>
+                                    <!-- REMOVED for GraalVM native-image support: Guava remains as transitive dependency -->
+                                    <!-- <include>com.google.guava:guava</include> -->
                                     <include>org.yaml:snakeyaml:*</include>
 
                                     <!-- Transitive dependency of guava -->
-                                    <include>org.checkerframework:*</include>
+                                    <!-- REMOVED for GraalVM native-image support: transitive dependency of guava -->
+                                    <!-- <include>org.checkerframework:*</include> -->
 
                                     <!-- Transitive dependency of httpclient5. Package name is org.apache.commons -->
                                     <include>commons-codec:*</include>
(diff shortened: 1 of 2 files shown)


sanzmauro:
@alexcol23 bringing this conversation over here as well 
https://github.com/splitio/java-client/pull/614#issuecomment-3862122542

nmayorsplit:
Hi @alexcol23,
I've just released java-client with a non-shaded option. You can download the new version (4.18.3) and add a classifier option in your pom file, as shown in the example below:

```
<dependency>
            <version>4.18.3</version>
            <classifier>non-shaded</classifier>
  </dependency>
```
Please let me know if it works for you.
Thank you!

wtorres02:
Hi @nmayorsplit @sanzmauro @alexcol23  

**The issue we encountered**

Following up on the non-shaded release of 4.18.3 — we were able to migrate from our custom fork and wanted to share our findings in case they help other users or inform a future release.

After switching to java-client:4.18.3 with the non-shaded classifier, our GraalVM native image build failed with:

Discovered unresolved type during parsing: `io.split.storages.pluggable.domain.UserStorageWrapper Caused by: java.lang.ClassNotFoundException: pluggable.CustomStorageWrapper`

UserStorageWrapper (present in the non-shaded jar) implements pluggable.CustomStorageWrapper, but that interface is not bundled in the non-shaded jar and is not declared as a transitive Maven dependency either. GraalVM's static analysis resolves the full type hierarchy at build time, so the missing interface causes a hard compilation failure — even if the pluggable storage feature is never used at runtime.

**How we fixed it**

We added `io.split.client:pluggable-storage` as an explicit dependency alongside the non-shaded client. This resolves the GraalVM error and the native image builds successfully.

**The concern**

The latest available version of pluggable-storage on Maven Central is 4.13.0-rc3, which is a release candidate. Shipping an RC artifact as a runtime dependency in production is not ideal, and we'd like to understand the roadmap here.

Would it be possible to either:
  1. Publish a stable GA release of pluggable-storage, or
  2. Add pluggable-storage as an explicit <optional>true</optional> Maven dependency on the non-shaded jar so it resolves transitively without users needing to declare it manually?

Happy to provide more context or a reproducer if useful. Thanks for the quick turnaround on the non-shaded release!

PR #618, opened 2026-02-06
[FME-12892] Add new non-shaded jar 

nmayorsplit:


.github/workflows/ci.yml
```diff
@@ -71,6 +71,10 @@ jobs:
 
       - name: SonarQube Scan (Pull Request)
         if: matrix.jdk == '11' && github.event_name == 'pull_request'
+        env:
+          SONARQUBE_HOST_URL: ${{ vars.SONARQUBE_HOST }}
+          SONAR_TOKEN: ${{ secrets.SONARQUBE_TOKEN }}
```
Israphel: already defined at upper level

these 3 variables don't go here, they already on a higher level because they're used for more than one step.

nmayorsplit: Thank you, I removed them.

melinaurruchua: you need to replace the old ones with the new ones

.github/workflows/ci.yml
```diff
@@ -30,6 +30,7 @@ jobs:
           - '11'
           - '19'
     env:
+      SONARQUBE_HOST_URL: ${{ vars.SONARQUBE_HOST }}
```
Israphel, line 33: artifactory variables below are not needed, we don't have artifactory.

client/pom.xml
```diff
@@ -257,4 +271,4 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
-</project>
+</project>
```
Israphel: broken end of file here

.github/workflows/ci.yml
nmayorsplit: If I remove it, the CI throws an error:
Screenshot 2026-02-26 at 4 46 23 PM
Screenshot 2026-02-26 at 4 46 46 PM

client/pom.xml
nmayorsplit: Updated.

gthea (approved):

